PR #20, opened 2017-04-01
woe to anyone who doesnt use 2.13 of gradle

epugh:
I seem to have picked up some fixing of line endings as well....


epugh:
Going to use Gradle wrapper instead.

PR #21, opened 2017-04-01
Use Gradle Wrapper to download very specific 2.13 version of gradle + Get CircleCI build to work + Fix RankyMcRankface dependency

epugh:
This is a better way of dealing with versions of Gradle!


softwaredoug:
Great PR, probably just want gradle to run `clean check` not just `test`

circle.yml
```diff
@@ -0,0 +1,7 @@
+machine:
+  java:
+    version: oraclejdk8
+
+test:
+  override:
+    - ./gradlew clean check
```
epugh: Looks good to me!   We may want to "collect" the unit tests etc in CircleCI at some point.

PR #27, opened 2017-04-14
Use ObjectParser and minor optimizations

nomoa:
- Use ObjectParser to simplify parsing code
- Always build ArrayList with the expected size to reduce gc pressure

### DIFF
--- a/src/main/java/com/o19s/es/ltr/query/LtrQueryBuilder.java
+++ b/src/main/java/com/o19s/es/ltr/query/LtrQueryBuilder.java
@@ -18,48 +18,53 @@
 package com.o19s.es.ltr.query;
 
 import ciir.umass.edu.learning.Ranker;
-import ciir.umass.edu.learning.RankerFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/com/o19s/es/ltr/query/LtrQueryBuilder.java
```diff
@@ -18,48 +18,53 @@
...
 public class LtrQueryBuilder extends AbstractQueryBuilder<LtrQueryBuilder> {
     public static final String NAME = "ltr";
+    private static final ObjectParser<LtrQueryBuilder, QueryParseContext> PARSER;
 
     Script _rankLibScript;
     List<QueryBuilder> _features;
-    String initialModel = null;
+
+    static {
+        PARSER = new ObjectParser<>(NAME, LtrQueryBuilder::new);
```
softwaredoug: I did not know this existed and I'm very happy I now know about it!

PR #29, opened 2017-04-26
Introduce Feature interfaces and reduce dependency on Ranklib

nomoa:
- Refactor the lucene query LtrQuery into RankerQuery
- Introduces Feature interfaces with a Prebuilt implemention used by the
  existing LtrQueryBuilder, strategy here will be to implement a Stored version
  of these interfaces:

  - Feature: a feature that can be transformed into a lucene query
  - FeatureSet: a container that holds a set of feature
  - LtrModel: a model that contains a ranker and a FeatureSet

- Make RankerQuery ranker agnostic with the LtrRanker interface
- Added a simple LinearRanker and a naive additive decision tree ranker
- Make RankerQuery a bit more lucene compliant (still need to figure out the
  impact on existing models that uses a Similarity that implements queryNorm)
- LtrQueryBuilder now creates a RankingQuery with a PrebuiltModel based on
  RanklibRanker

The purpose is to start to reduce the dependency on Ranklib and prepare future
developpement to implement a feature store.


nomoa:
@ebernhardson, @softwaredoug  this is a preliminary patch in preparation of the feature store.
This patch adds no new features to the plugin and should remain back-compatible (except for queryNorm and feature logging with msearch which might be inconsistent when used with the classic similarity).
The most important thing here I think is the new interfaces Feature/FeatureSet/LtrModel which should be the base for the feature store.
I've added two new rankers (not yet used) just to be sure that RankerQuery was flexible enough to support multiple ranker.

src/main/java/com/o19s/es/ltr/feature/PrebuiltFeatureSet.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public String name() {
+        return name;
+    }
+
+    @Override
+    public List<Feature> features() {
+        return Arrays.asList(features);
```
gehel: The returned list should probably at least be unmodifiable. I'm not sure what dependencies are available, but if there is some kind of immutable list available, it would probably make sense to use it. If that immutable list is available, it should be used instead of an array as an instance member.

src/main/java/com/o19s/es/ltr/query/LtrQueryBuilder.java
```diff
@@ -114,17 +119,19 @@ protected Query doToQuery(QueryShardContext context) throws IOException {
         if (_features == null || _rankLibScript == null) {
             return new MatchAllDocsQuery();
         }
-        List<String> featureNames = new ArrayList<String>(_features.size());
-        List<Query> asLQueries = new ArrayList<Query>(_features.size());
-        for (QueryBuilder query : _features) {
-            asLQueries.add(query.toQuery(context));
-            featureNames.add(query.queryName());
+        PrebuiltFeature[] features = new PrebuiltFeature[_features.size()];
+        for(int i = 0; i < _features.size(); i++) {
```
gehel: This looks to me like something that would be more readable as a stream.map().

src/main/java/com/o19s/es/ltr/ranker/ranklib/DenseProgramaticDataPoint.java
```diff
@@ -14,16 +14,17 @@
...
 
 /**
  * Implements DataPoint but without needing to pass in a stirng
  * to be parsed
  */
-public class DenseProgramaticDataPoint extends DataPoint {
+public class DenseProgramaticDataPoint extends DataPoint implements LtrRanker.DataPoint {
 
     public DenseProgramaticDataPoint(int numFeatures) {
         super();
```
gehel: The call to super() is implicit, it could be removed (minor style issue)

src/main/java/com/o19s/es/ltr/ranker/ranklib/DenseProgramaticDataPoint.java
```diff
@@ -53,4 +54,14 @@ public void setFeatureVector(float[] dfVals) {
     public float[] getFeatureVector() {
         return this.fVals;
     }
+
+    @Override
+    public void setFeatureScore(int featureIdx, float score) {
+        this.setFeatureValue(featureIdx+1, score);
```
gehel: Why the `+1`? An inline comment explaining it would be welcomed.

gehel: Oh, I see that this list is 1-based... still, the comment would be welcomed as this is unusual enough.

src/main/java/com/o19s/es/ltr/feature/FeatureSet.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    String name();
+
+    /**
+     * List of feature present in this model.
+     * This list must be properly ordered to that {@link #featureOrdinal(String)}
+     * matches features().get(ordinal)
+     * @return the list of features in the set.
+     */
+    List<Feature> features();
```
nomoa: right, reading back this interface I'm not super happy with it. I'll simplify a bit.
I'm anticipating some future usage and it's likely to change when I'll have to implement real stuff on top of it.

src/main/java/com/o19s/es/ltr/feature/LtrModel.java
```diff
@@ -12,22 +12,31 @@
...
 
 /**
- * Created by doug on 12/23/16.
+ * Represents a self contained LTR model
  */
-public class RankLibQuery extends CustomScoreQuery {
+public interface LtrModel {
+    /**
+     * Name of the model
+     * @return the name
```
nomoa: Sure, in fact I can remove some comment lines, javadoc complained about the empty `@return` line but is happy if I remove it completely.

src/main/java/com/o19s/es/ltr/feature/PrebuiltFeature.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * A prebuilt featured query, needed by query builders
+ * that provides the query itself.
+ */
+public class PrebuiltFeature implements Feature {
+    private final String name;
+    private final Query query;
+
+    public PrebuiltFeature(String name, Query query) {
+        this.name = name;
```
nomoa: Yes for now we still have to support null names, adding `@Nullable` from elastic

src/main/java/com/o19s/es/ltr/feature/PrebuiltFeatureSet.java
nomoa: Removed, added a asArray() method that I'll use internally

src/main/java/com/o19s/es/ltr/feature/PrebuiltFeatureSet.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public int featureOrdinal(String featureName) {
+        // slow, not meant for prod usage
+        for (int i = 0; i < features.length; i++) {
+            if(Objects.equals(features[i].getName(), featureName)) {
```
nomoa: done

src/main/java/com/o19s/es/ltr/query/LtrQueryBuilder.java
nomoa: Sadly I can't, toQuery(context) throws IOException :(

src/main/java/com/o19s/es/ltr/ranker/ranklib/DenseProgramaticDataPoint.java
nomoa: done

src/main/java/com/o19s/es/ltr/ranker/ranklib/DenseProgramaticDataPoint.java
nomoa: done

nomoa:
Thanks for the review!
Pushing a new patch with your comments and some cleanups.
I'll squash them if you tell me it's ok.

src/main/java/com/o19s/es/ltr/query/LtrQueryBuilder.java
gehel: There are some ways to get around that ([CompletableFutures](http://stackoverflow.com/questions/30117134/aggregate-runtime-exceptions-in-java-8-streams) come to mind), but it is probably not worth it. It might make sense to invest in streams if we can move enough of the transformations there as it would allow us to get rid of a lot of intermediate collections. That being said, this should not be a priority at this point.

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -201,30 +206,32 @@ public void normalize(float norm, float boost) {
...
+             * NOTE: Switch to ChildScorer and {@link #getChildren()} if it appears
+             * to be useful for logging
+             */
+            private final ArrayList<Scorer> scorers;
             private final DocIdSetIterator iterator;
-            private final float[] scores;
             private final LtrRanker.DataPoint dataPoint;
 
-            RankerScorer(RankerChildScorer[] scorers, DocIdSetIterator iterator) {
+            RankerScorer(ArrayList<Scorer> scorers, DocIdSetIterator iterator) {
```
gehel: To respect DI, you probably want a List<Scorer> and not an ArrayList<Scorer>. Unless you really want an ArrayList (which might well be the case for performance reasons). The actual constraint is probably List & RandomAccess, but expressing that constraint is fairly ugly.

In short: you probably need to ignore the comment above :)

src/main/java/com/o19s/es/ltr/feature/PrebuiltFeatureSet.java
```diff
@@ -35,34 +37,46 @@ public String name() {
     }
 
     @Override
-    public List<Feature> features() {
-        return Arrays.asList(features);
+    public PrebuiltFeature[] asArray() {
+        return features.toArray(new PrebuiltFeature[features.size()]);
```
gehel: Copying a list to an array can get expansive. I wonder if there is a reason to have a different internal vs external representation of features (except to protect the internal state as it is mutable).

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
nomoa: Yes... that's the problem, it hurts my eyes as well to see a var defined as `ArrayList<Type> var = ...`
But yes RandomAccess is critical here, and not sure what to choose between:
`Scorer[]` vs `ArrayList<Scorer>`
In the end what's important is random access and the fact that the array that holds the data is created once (not auto adjusted when building with things like ArrayList#ensureCapacity)

src/main/java/com/o19s/es/ltr/feature/PrebuiltFeatureSet.java
nomoa: I really don't know yet how it'll look like for other implementations...
This class has been created to start using the new model with the existing query builder. 
This will likely change in the future.

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -0,0 +1,320 @@
...
+             */
+            private final ArrayList<Scorer> scorers;
+            private final DocIdSetIterator iterator;
+            private final LtrRanker.DataPoint dataPoint;
+
+            RankerScorer(ArrayList<Scorer> scorers, DocIdSetIterator iterator) {
+                super(RankerWeight.this);
+                this.scorers = scorers;
+                this.iterator = iterator;
+                dataPoint = ranker.newDataPoint();
```
nomoa: makes sense, will adjust

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -0,0 +1,320 @@
...
+            }
+
+            @Override
+            public float score() throws IOException {
+                for (int i = 0; i < scorers.size(); i++) {
+                    Scorer scorer = scorers.get(i);
+                    if (scorer.docID() == docID()) {
+                        dataPoint.setFeatureScore(i, scorer.score());
+                    } else {
+                        dataPoint.setFeatureScore(i, 0);
```
nomoa: Sure, then I'll probably defer this to the feature vector and implement a reset() method and remove the else part or a `reuse` param as seen sometimes in lucene...

nomoa: Also I'm struggling to visualize how a sparse lucene query would look like, today it's clear that we make the bold assumption that the model is dense:
`initialFeatures.length == queries.length == scorers.length == featureVector.length >= numberOfDistinctFeatureUsedInTheModel`
and that we can use the array index as a feature identifier.
If we go with a sparse approach this equality will break somewhere and we will  need some indirection to infer the featureIdentifier.

Looking at the solr implementation I see that it's relatively dense as well https://github.com/apache/lucene-solr/blob/master/solr/contrib/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java#L406 despite having a strong distinction between Sparse and Dense scorers. For solr the feature remains identified by its index in the array.

I agree that we currently lack a way to describe missing values.

If dense vs sparse is just something to allow identifying missing values then it should be relatively easy to do.
On the other hand if sparse means that the equality I wrote earlier is not verified then I think we need more examples to clarify its usage.

gehel:
I much prefer this version with just collections. The rewrite of List to ArrayList conditional on RandomAccess seems to be a nice compromise. Still not immutable, but it is probably reasonable to extend some trust to client classes.

sstults:
@nomoa I created a 1_0 branch off of master. Should we change this to a PR for it?

PR #43, opened 2017-05-12
Implement base components for the feature store

### LINKED ISSUE #37: Execute LTR queries without specifying features
- Pass a model
- Needs to be parameterized to include whatever the templates require
  - Query string
  - Location
  - ...

nomoa:
Implement the stored version of the feature/set/model interfaces with their
json parsing logic.
Extracted LtrRanker parsing logic into a LtrRankerParser interface.
Implemented a CachedFeatureStore wrapper.
Added various test utils to generate random features, sets and models.
Added a StoredLtrQueryBuiler implementation that takes a model name
and a set of parameters.
Had to mockup mustache with a naive template engine because mustache is not
available during AbstractQueryBuilder tests.
Hopefully this is last giant patch, following patches should be smaller:
- add more LtrRankerParser implementations
- add a REST api on top the IndexFeatureStore
- add a logging API

closes #37

### DIFF
--- a/src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
+++ b/src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
@@ -16,30 +16,66 @@
  */
 package com.o19s.es.ltr;
 
+import ciir.umass.edu.learning.RankerFactory;
+import com.o19s.es.ltr.feature.store.index.Caches;
 import com.o19s.es.ltr.query.LtrQueryBuilder;
+import com.o19s.es.ltr.query.StoredLtrQueryBuilder;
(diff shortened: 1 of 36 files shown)


nomoa:
@sstults, @ebernhardson, @gehel sorry again for the giant patch but this should be the last one.
I needed to put a lot of classes in place in order to be able to have all the contexts needed to run a ltr query without specifying the features.

Let me know if you want to walk through it together.

Thanks for your help and reviews!

softwaredoug:
@nomoa when would be a good time to go through this PR?

softwaredoug:
Oh this is into 1.0, right now I'm generally ok with accepting stuff liberally into 1.0 with the understanding we might back it out later or change our approach some other time.

PR #44, opened 2017-05-13
Bumped to ES 5.4.0. Appears to require Gradle upgrade to build

sstults:


softwaredoug:
Hmm I think circle CI works usually, but I bet the gradle upgrade broke something. @epugh might have some ideas.

sstults:
I added @epugh as a reviewer. May be anecdotal, but the build cleared CI after bumping the Gradle version in gradle-wrapper.properties.

PR #45, opened 2017-05-15
Fix the problem of parsing sample_judgements.txt in the demo

BoeingX:
The one-pass parsing of `sample_judgements.txt` skips the first judgement. Indeed, the function `_queriesFromHeader` breaks when it meets the first line not beginning with `#`. When the file handler is passed to `_judgmentsFromBody`, the first judgement is inevitably skipped. While it is possible to solve the problem by adding a blank line between the header and the body, the two-pass version is less error-prone.


softwaredoug:
Thanks!

PR #48, opened 2017-05-30
WIP Json serialization

softwaredoug:
First pass at specifying models via JSON  for 1.0. This should be a big step towards breaking ranklib dependency. Currently supports mart/linear models. Should be easy to add Random Forest models.


softwaredoug:
Delaying until 1.1 or later

PR #49, opened 2017-06-01
Add a REST layer on top of the feature store

### LINKED ISSUE #33: Feature Store API: Add ability to assign unique name to features

### LINKED ISSUE #34: Feature Store API: Add ability to group features as feature sets

nomoa:
Added the following:
- Simple create/delete operations for manipulating stores.
- Simple CRUD operations for feature, featureset and model.
- Clear cache operation
- Fetch cluster wide cache stats
- Create/update a featureset from existing features
- Create a model from an existing featureset

Stores are not created automatically and must be created before using them.
Features and featureset can be updated, models can only be created.

closes #34
closes #33


nomoa:
@ebernhardson, @gehel a new giant patch to review, sorry! :(

src/main/java/com/o19s/es/ltr/action/TransportAddFeatureToSetAction.java
```diff
@@ -0,0 +1,243 @@
...
+                }
+            } catch (Exception e) {
+                onActionFailure(e);
+            }
+        }
+
+        private void onActionFailure(Exception e) {
+            exception.set(e);
+            if (countdown.countDown()) {
+                listener.onFailure(exception.get());
```
gehel: Here, exception.get() != e (at least some of the time if both search and get are in error). This is probably a mistake, but I'm not entirely sure of the context.

src/main/java/com/o19s/es/ltr/action/TransportAddFeatureToSetAction.java
```diff
@@ -0,0 +1,243 @@
...
+            }
+        }
+
+        private void mergeAndStore() {
+            if (exception.get() != null) {
+                listener.onFailure(exception.get());
```
gehel: Don't we notify the listener twice of that exception? Once in onActionFailure(), once here? Also, what happens if there is more than one exception?

src/main/java/com/o19s/es/ltr/action/TransportClearCachesAction.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    protected ClearCachesNodeResponse newNodeResponse() {
+        return new ClearCachesNodeResponse();
+    }
+
+    @Override
+    protected ClearCachesNodeResponse nodeOperation(ClearCachesNodeRequest request) {
+        ClearCachesNodesRequest r = request.request;
+        switch (r.getOperation()) {
+        case ClearStore: {
```
gehel: Code blocks in switch statements? Syntactically correct, but I know only one person who uses code blocks... (well, now two).

src/main/java/com/o19s/es/ltr/action/TransportFeatureStoreAction.java
```diff
@@ -0,0 +1,138 @@
...
+                indexNameExpressionResolver, FeatureStoreRequest::new);
+        this.factory = factory;
+        this.clusterService = clusterService;
+        this.clearCachesAction = clearCachesAction;
+        this.client = client;
+    }
+
+    @Override
+    protected void doExecute(FeatureStoreRequest request, ActionListener<FeatureStoreResponse> listener) {
+        logger.warn("attempt to execute a TransportFeatureStoreAction without a task");
```
gehel: logging AND throwing exception is usually redundant. I would expect the caller to log exception as well, so we will probably end up with duplicate log messages (well, the messages will be slightly different, but the point is that we could have all the context in the exception and remove the logger.warn() call). That same pattern is used in a few other places.

docs/rest_api.md
```diff
@@ -0,0 +1,147 @@
...
+
+Feature names in the set must be unique.
+
+To use a custom store simply prefix with the name of the store:
+`/_ltr/custom_store/_feature/feature_name`
+
+### Append features to a set
+A set can also be created/updated from existing features:
+
+`POST /_ltr/_featureset/my_featureset/_addfeatures/my_feature*`
```
gehel: After discussion with @nomoa, this operation actually fails the second time as the features have already been added. POST make sense.

:+1:

src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
```diff
@@ -17,40 +17,85 @@
...
 
     public LtrQueryParserPlugin(Settings settings) {
         caches = new Caches(settings);
-        parserFactory = new LtrRankerParserFactory.Builder()
-                .register(RanklibModelParser.TYPE, () -> new RanklibModelParser(getRanklibFactory()))
+        parserFactory = buildParserFactory();
+    }
+
+    public static LtrRankerParserFactory buildParserFactory() {
+        Supplier<RankerFactory> ranklib = Suppliers.memoize(RankerFactory::new);
```
nomoa: added a comment

src/main/java/com/o19s/es/ltr/action/TransportAddFeatureToSetAction.java
```diff
@@ -0,0 +1,243 @@
...
+                nameQuery = QueryBuilders.matchQuery("name", featureNamesQuery);
+            }
+            BoolQueryBuilder bq = QueryBuilders.boolQuery();
+            bq.must(nameQuery);
+            bq.must(QueryBuilders.matchQuery("type", StoredFeature.TYPE));
+            srequest.types(IndexFeatureStore.ES_TYPE);
+            srequest.source().query(bq);
+            srequest.source().fetchSource(true);
+            srequest.source().size(StoredFeatureSet.MAX_FEATURES);
+            searchAction.execute(srequest, wrap(this::onSearchResponse, this::onActionFailure));
```
nomoa: done

src/main/java/com/o19s/es/ltr/action/TransportAddFeatureToSetAction.java
nomoa: refactored

src/main/java/com/o19s/es/ltr/action/TransportAddFeatureToSetAction.java
nomoa: refactored

src/main/java/com/o19s/es/ltr/action/TransportClearCachesAction.java
nomoa: done

src/main/java/com/o19s/es/ltr/action/TransportFeatureStoreAction.java
nomoa: done

src/main/java/com/o19s/es/ltr/action/CachesStatsAction.java
```diff
@@ -0,0 +1,246 @@
...
+        @Override
+        public void writeTo(StreamOutput out) throws IOException {
+            super.writeTo(out);
+            all.writeTo(out);
+        }
+
+        @Override
+        public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+            builder.field("all", all);
+            builder.startArray("nodes");
```
nomoa: makes sense, fixing

ebernhardson:
I was able to load in some features, create a feature set, add an xgboost model (using followup patch), and run an "sltr" query against it. According to the explain this generated appropriate feature vectors, although for some reason all my results get the same score, but this might simply be because the model was trained against a full enwiki index and i have a tiny index of 200 documents locally, so the split values might all be useless.

ebernhardson:
I was able to load in a dozen features, build a feature set out of them, and then with the xgboost json followup patch load in an xgboost model. I was then able to execute an "sltr" query and get results, although ever result has the same score. This is very likely because the index trained against and the index i'm testing against are completely different, so probably nothing to worry about yet.

ebernhardson:
Above problem with same values for everything was indeed just an unrepresentative index. I pulled in 100k docs from one of the dumps and i now get varied scores for documents.

ebernhardson:
It might also be nice to have some basic information on a few endpoints that are currently unclaimed:

GET /_ltr - Perhaps a list of stores available? not sure
GET /_ltr/_feature - List of features?
GET /_ltr/_featureset - List of named feature sets?
GET /_ltr/_model - List of models?

src/main/java/com/o19s/es/ltr/action/TransportCreateModelFromSetAction.java
```diff
@@ -0,0 +1,102 @@
...
+                .type(IndexFeatureStore.ES_TYPE)
+                .id(StorableElement.generateId(StoredFeatureSet.TYPE, request.getFeatureSetName()));
+        getRequest.setParentTask(clusterService.localNode().getId(), task.getId());
+        getAction.execute(getRequest, ActionListener.wrap((r) -> this.doStore(task, r, request, listener), listener::onFailure));
+    }
+
+    private void doStore(Task parentTask, GetResponse response, CreateModelFromSetRequest request,
+                         ActionListener<CreateModelFromSetResponse> listener) {
+        if (!response.isExists()) {
+            throw new IllegalArgumentException("Stored feature [" + request.getFeatureSetName() + "] does not exist");
```
nomoa: done

src/main/java/com/o19s/es/ltr/action/TransportCreateModelFromSetAction.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        if (request.getExpectedSetVersion() != null && request.getExpectedSetVersion() != response.getVersion()) {
+            throw new IllegalArgumentException("Stored feature [" + request.getFeatureSetName() + "]" +
```
nomoa: done

docs/rest_api.md
```diff
@@ -0,0 +1,147 @@
...
+
+Will create or update the `my_featureset` by collecting all features named with the prefix `my_feature`.
+
+### Models
+
+`/_ltr/_model/feature_name` supports `PUT`, `GET` and `DELETE` operations.
```
nomoa: done

docs/rest_api.md
```diff
@@ -0,0 +1,147 @@
...
+Clear the cache for a feature store:
+```
+POST /_ltr/_clearcache
+POST /_ltr/custom_store/_clearcache
+```
+
+Get cluster wide cache statistics:
+```
+GET /_ltr/_cachestats
+```
```
nomoa: Added a small note under the Feature Store section

src/main/java/com/o19s/es/ltr/action/FeatureStoreAction.java
```diff
@@ -0,0 +1,216 @@
...
+                arve = addValidationError("store must be set", null);
+            }
+            if (!store.equals(IndexFeatureStore.DEFAULT_STORE) && !store.startsWith(IndexFeatureStore.STORE_PREFIX)) {
+                arve = addValidationError("Store name [" + store + "] is invalid.", arve);
+            }
+            if (storableElement == null) {
+                arve = addValidationError("storable element must be set", arve);
+            }
+            if (action == Action.UPDATE && !storableElement.updatable()) {
+                arve = addValidationError("Element of type [" + storableElement.type() + "] are not updatable.", arve);
```
nomoa: done

src/main/java/com/o19s/es/ltr/rest/RestCreateModelFromSet.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.IOException;
+
+public class RestCreateModelFromSet extends FeatureStoreBaseRestHandler {
+    private static final ObjectParser<ParserState, Void> PARSER;
+    static {
+        PARSER = new ObjectParser<>("model", ParserState::new);
+        PARSER.declareString(ParserState::setName, new ParseField("name"));
+        PARSER.declareString(ParserState::setModelType, new ParseField("type"));
+        PARSER.declareString(ParserState::setModelDefinition, new ParseField("definition"));
```
nomoa: makes sense, not sure how to do that cleanly yet.
I'll try to change the REST api for this and use the Content-Type http header to read the model name.
Doing this while parsing is bit tricky since I'd need to change the model representation from String to either String/XContent ...

src/main/java/com/o19s/es/ltr/action/ClearCachesAction.java
```diff
@@ -0,0 +1,177 @@
...
+            ActionRequestValidationException arve = null;
+            if (store == null) {
+                arve = addValidationError("store cannot be null", null);
+            }
+
+            if (operation == null) {
+                arve = addValidationError("no operation provided", arve);
+            }
+
+            if (operation != null && operation != Operation.ClearStore && name == null) {
```
nomoa: in fact it can, we do not fail on the previous if and here we continue to collect validation errors so we must check against null not to add a second error.

nomoa:
@ebernhardson I think I addressed all your comments, in the end I did like you suggested concerning json models: I do exactly the same as I do for feature with a `modelAsString` flag.
I added a simple LinearRankerParser for easy testing.

@gehel if you have 5 mins I'd love if you could check the last 3 patches, thanks! :)

src/main/java/com/o19s/es/ltr/rest/RestSimpleFeatureStore.java
```diff
@@ -0,0 +1,249 @@
...
+                });
+    }
+
+    RestChannelConsumer search(NodeClient client, String type, String indexName, RestRequest request) {
+        String prefix = request.param("prefix");
+        int from = request.paramAsInt("from", 0);
+        int size = request.paramAsInt("size", 20);
+        BoolQueryBuilder qb = QueryBuilders.boolQuery().filter(QueryBuilders.termQuery("type", type));
+        if (prefix != null && !prefix.isEmpty()) {
+            qb.must(QueryBuilders.matchQuery("name.prefix", prefix));
```
gehel: A static import of QueryBuilders.matchQuery would make this read more fluently (very minor, feel free to ignore).

src/main/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStore.java
```diff
@@ -53,44 +92,170 @@ public String getStoreName() {
...
-        return getAndParse(id, FEATURE_SET_TYPE, StoredFeatureSet::parse);
+    public StoredFeatureSet loadSet(String name) throws IOException {
+        return getAndParse(name, StoredFeatureSet.class, StoredFeatureSet.TYPE);
+    }
+
+    /**
+     * Construct the elasticsearch index name based on a store name
+     */
+    public static String indexName(String storeName) {
+        if (Objects.requireNonNull(storeName).isEmpty()) {
```
gehel: We throw a NPE with no message if `storeName` is `null` but an `IllegalArgumentException` if it is empty. This sounds like a job for Guava's `Preconditions.checkArgument()` and `Strings.isNullOrEmpty()`, but we don't have it as a dependency yet (probably not worth adding it, so feel free to ignore).

src/test/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStoreTests.java
```diff
@@ -0,0 +1,124 @@
...
+                randomSimpleString(random(), 5, 10),
+                randomRealisticUnicodeString(random(), 5, 1000),
+                true));
+    }
+
+    public void testIsIndexName() {
+        assertTrue(isIndexStore(IndexFeatureStore.DEFAULT_STORE));
+        assertFalse(isIndexStore("not_really"));
+        assertFalse(isIndexStore(IndexFeatureStore.STORE_PREFIX));
+        for (int i = 0; i < random().nextInt(10) + 10; i++) {
```
gehel: `random().nextInt(10) + 10` is evaluated for each step of the loop. It is technically correct, but it feels a bit strange to have a non constant bound to a loop (feel free to ignore).

Same comment applies to `testStoreName()`.

src/main/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStore.java
nomoa: I'll ignore that one :)

src/test/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStoreTests.java
nomoa: doh I'm stupid, fixing...

ebernhardson:
I don't know if it needs to go here, this patch is big enough, but i've been trying to debug getting a no match for all templated features and it might be useful to have some sort of debug endpoint that can report back what the query generated for a feature is.

src/main/java/com/o19s/es/ltr/ranker/parser/XGBoostJsonParser.java
```diff
@@ -0,0 +1,168 @@
...
+                    (missingNodeId == null || missingNodeId.equals(children.get(1).nodeId));
+        }
+        boolean isSplit() {
+            return leaf == null;
+        }
+
+
+        Node toNode(FeatureSet set) {
+            if (isSplit()) {
+                return new NaiveAdditiveDecisionTree.Split(children.get(1).toNode(set), children.get(0).toNode(set),
```
nomoa: sorry!!
reading back your comments on my fork I completely missed your first line about this issue...
Fixing...
I think I'm confused by the meaning of a successful condition:
for me a condition is met if `feature_score > split.threshold` => `right` but it's `feature_score < split.threshold` => `left`.
So for me `yes` was `go right`...
We don't really care but `missing` => `no` while it would make more sense to go to `yes`?
Well... I'll ignore completely `missing`... I have added some code to constraint the structure but since we don't use it it's confusing to enforce a particular side for `missing`.

gehel:
From my limited understanding, those last 2 commits look good to me...

nomoa:
concerning missing params I default to `match_none` if one of them is missing.
That is probably misleading, I could let the user specify a default query to use if missing and fail if no default is set?

nomoa:
NOTE: a version of the plugin using this branch is being deployed to our production clusters for load testing and possibly a/b test.
merging so that it'll be easier to track future changes.

PR #50, opened 2017-06-20
New stat scoring features

worleydl:


src/main/java/com/o19s/es/explore/ExplorerFetchSubPhase.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Set;
+
+/**
+ * Created by Daniel on 6/15/2017.
+ */
+public class ExplorerFetchSubPhase implements FetchSubPhase {
+    @Override
+    public void hitExecute(SearchContext context, HitContext hitContext) {
+        if (hitContext.hit().fieldsOrNull() == null) {
+            hitContext.hit().fields(new HashMap<>(2));
```
gehel: With an initial capacity of 2 and a default load factor of .75, this hashmap will most probably get rehased sooner than you think. Have a look at [how Guava does it](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Maps.java#L274-L307).

src/main/java/com/o19s/es/explore/StatisticsHelper.java
```diff
@@ -0,0 +1,87 @@
...
+    float getStdDev() {
+        return (float) Math.sqrt(getVariance());
+    }
+
+    public float median() {
+        Integer[] sortedData = data.toArray(new Integer[data.size()]);
+        Arrays.sort(sortedData);
+
+        if (sortedData.length % 2 == 0)
+        {
```
gehel: minor style issue: this seems to be the only place when curly brackets are on a new line...

src/main/java/com/o19s/es/explore/StatisticsHelper.java
```diff
@@ -0,0 +1,87 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.o19s.es.explore;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class StatisticsHelper {
```
nomoa: I think you need to protect against an edge case where there are no points in the data array.
Or add `assert !data.isEmpty();` before all access to data to indicate future devs that this class cannot be used with an empty data array.

src/main/java/com/o19s/es/explore/ExplorerFetchSubPhase.java
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            Query query = context.query();
+            Weight weight = query.createWeight(context.searcher(), false);
+            Set<Term> terms = new HashSet<Term>();
+            weight.extractTerms(terms);
+
+            // No stats if no terms
+            if(terms.size() == 0) return;
+
+            Terms termVector = hitContext.reader().getTermVector(hitContext.docId(), extBuilder.getField());
```
nomoa: (forgot to add this comment in my previous review, sorry)

I can see 2 drawbacks of using the termVector these kind of stats:
- it's rather slow
- it's not always available

In our context we do not index them and we might have to hack our mappings and reindex to take benefit of your new query (@ebernhardson might have some opinion on this?).
We could generate termVectors on the fly but this would be even more slow and complex.

I'm not strongly against using termVectors for a first iteration to decide if they are useful but I'm not too keen running this kind of query in production, it'll fetch a rather large portion of data from disk on a lot of docs.

Also I don't think it's very complex to extract the stats we want without using termVectors.
First thing is to do is to distinguish the scope of stat source:
- Collection scoped: the score is the the same for all docs
  - `raw_df`: `TermStatistics#docFreq()`
  - `raw_ttf`: `TermStatistics#sumTotalFreq()`
- Doc scoped: different on every doc
  - `raw_tf`: `PostingEnum#freq()`: 
    - tip: here you use subscorer and I think you can easily fetch subterms freq by iterating on `subscorer.getChildren()`:
```java
        for (ChildScorer child : subscorer.getChildren()) {
            if (child.child.docID() == docID()) {
                tf_stats.add(child.child.freq());
            }
        }
```
`Scorer#getChildren()` highly depends on the nested query we pass, but for simple `match_query` it should work (not tested).

src/main/java/com/o19s/es/explore/ExplorerScorer.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public float score() throws IOException {
+        // TODO: Cleanup DRY violation, similar code in the Explorer fetch phase
+        Set<Term> terms = new HashSet<Term>();
+        weight.extractTerms(terms);
+
+        // No stats if no terms
+        if(terms.size() == 0) return 0.0f;
+
+        Terms termVector = context.reader().getTermVector(docID(), field);
```
nomoa: (forgot to add this comment in my previous review, sorry)

I can see 2 drawbacks of using the termVector for these kind of stats:
- it's rather slow
- it's not always available

In our context we do not index them and we might have to hack our mappings and reindex to take benefit of your new query (@ebernhardson might have some opinion on this?).
We could generate termVectors on the fly but this would be even more slow and complex.

I'm not strongly against using termVectors for a first iteration (to decide if these features are useful) but I'm not too keen running this kind of query in production, it'll fetch a rather large portion of data from disk on a lot of docs.

Also I don't think it's very complex to extract the stats we want without using termVectors.
First thing is to do is to distinguish the scope of stat source:
- Collection scoped: the score is the the same for all docs a constant scorer might be sufficient
  - `raw_df`: `TermStatistics#docFreq()`
  - `raw_ttf`: `TermStatistics#sumTotalFreq()`
- Doc scoped: different on every doc, a custom scorer like this one is needed
  - `raw_tf`: `PostingEnum#freq()`: 
    - tip: here you use subscorer and I think you can easily fetch subterms freq by iterating on `subscorer.getChildren()`:
```java
        for (ChildScorer child : subscorer.getChildren()) {
            if (child.child.docID() == docID()) {
                tf_stats.add(child.child.freq());
            }
        }
```
`Scorer#getChildren()` highly depends on the nested query we pass, but for simple `match_query` it should work (not tested).

src/test/java/com/o19s/es/explore/ExplorerQueryTests.java
```diff
@@ -0,0 +1,92 @@
...
+        IndexWriter indexWriter = new IndexWriter(dir, new IndexWriterConfig(Lucene.STANDARD_ANALYZER));
+
+        for(int i = 0; i < docs.length; i++) {
+            Document doc = new Document();
+            doc.add(new Field("_id", Integer.toString(i + 1), StoredField.TYPE));
+            doc.add(newTextField("text", docs[i], Field.Store.YES));
+            indexWriter.addDocument(doc);
+        }
+
+        indexWriter.close();
```
gehel: a try-with-resources would be cleaner than an explicit close

src/test/java/com/o19s/es/explore/ExplorerQueryTests.java
```diff
@@ -0,0 +1,92 @@
...
+
+        reader = DirectoryReader.open(dir);
+        searcher = new IndexSearcher(reader);
+    }
+
+    @After
+    public void cleanup() throws Exception {
+        reader.close();
+        dir.close();
```
gehel: there is a missing `try-finally` to ensure `dir` is closed even if `reader.close()` throws an exception (yes, I know, this is only a test class...

worleydl: No worries, good suggestion.  Also sorry this is still a bit of a work in progress, I should have held on pushing but thanks for the tips!

worleydl:
All feedback should be addressed now.

Notes:
- Removed median from stats helper since it wasn't being used by any of our features.
- Using scorer.freq() in place of term vectors doesn't allow us to specify a custom field anymore.  Instead the field parsed in the query is used.  I think this makes sense but if we need to be able to change the field I'll need to revisit.

src/test/java/com/o19s/es/explore/StatisticsHelperTests.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.lucene.util.LuceneTestCase;
+import org.junit.Test;
+
+public class StatisticsHelperTests extends LuceneTestCase {
+    private final float[] dataset = new float[] {
+      0.0f, -5.0f, 10.0f, 5.0f
+    };
+
+    @Test(expected=AssertionError.class)
```
nomoa: forbidden api will complain here, junit anotations are not allowed.
also you don't really need to test this, inline assertion are meant to be tested alongside normal tests and will blow-up if something went wrong.

src/main/java/com/o19s/es/explore/ExplorerScorer.java
```diff
@@ -0,0 +1,100 @@
...
+    protected ExplorerScorer(Weight weight, LeafReaderContext context, String type, Scorer subScorer) {
+        super(weight);
+        this.context = context;
+        this.type = type;
+        this.subScorer = subScorer;
+    }
+
+    @Override
+    public float score() throws IOException {
+        Set<Term> terms = new HashSet<Term>();
```
nomoa: I don't think you need to re-extract terms here, I think you can remove the terms var

src/main/java/com/o19s/es/explore/ExplorerQueryBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+        try {
+            builder = PARSER.parse(context.parser(), context);
+        } catch (IllegalArgumentException iae) {
+            throw new ParsingException(parser.getTokenLocation(), iae.getMessage(), iae);
+        }
+
+        if (builder.query == null) {
+            throw new ParsingException(parser.getTokenLocation(), "Field [" + QUERY_NAME + "] is mandatory.");
+        }
+        if (builder.statsType() == null) {
```
nomoa: for later: when we'll have n enum it'll be easy to check that stats type is invalid and throw a ParsingException here.

src/main/java/com/o19s/es/explore/ExplorerQueryBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+        out.writeString(type);
+    }
+
+    @Override
+    protected void doXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject(NAME);
+
+        printBoostAndQueryName(builder);
+
+        builder.field(QUERY_NAME.getPreferredName());
```
nomoa: nitpick: you can pack these 2 lines into one:
`builder.field(QUERY_NAME.getPreferredName(), query);`

src/main/java/com/o19s/es/explore/ExplorerQueryBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    protected void doWriteTo(StreamOutput out) throws IOException {
+        out.writeNamedWriteable(query);
+        out.writeString(type);
+    }
+
+    @Override
+    protected void doXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject(NAME);
+
```
nomoa: nitpick: the code would look better without the empty lines

src/main/java/com/o19s/es/explore/ExplorerQueryBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+public class ExplorerQueryBuilder extends AbstractQueryBuilder<ExplorerQueryBuilder> implements NamedWriteable{
+    public static final String NAME = "match_explorer";
+
+    private static final ParseField QUERY_NAME = new ParseField("query");
+    private static final ParseField TYPE_NAME = new ParseField("type");
+
+    private QueryBuilder query;
```
nomoa: for later: concerning the problem of not being able to specify the field:
I think it's ok for now, the solution I could think of (if working on top of another query is fragile) is:
instead of wrapping a query we could simply accept the text string and analyze it ourselves, we'll then have better control over the lucene query being created.

src/main/java/com/o19s/es/explore/StatisticsHelper.java
```diff
@@ -0,0 +1,87 @@
...
+        return getSum() / data.size();
+    }
+
+    public float getSum() {
+        assert !data.isEmpty();
+
+        float sum = 0.0f;
+
+        for(float a : data) {
+            sum += a;
```
nomoa: I have some concerns about overflows and precisions, some stats may be relatively large numbers and I worry that float is not well suited for this. Sadly float is the type used by lucene...
I don't have great solutions but this certainly something we'll have to address sooner or later, esp. if we want to combine features.

src/main/java/com/o19s/es/explore/ExplorerQuery.java
```diff
@@ -0,0 +1,243 @@
...
+    private final Query query;
+    private final String type;
+
+    public ExplorerQuery(Query query, String type) {
+        this.query = query;
+        this.type = type;
+    }
+
+    private boolean isCollectionScoped() {
+        return type.contains("_df")
```
nomoa: endsWith?

src/main/java/com/o19s/es/explore/ExplorerQuery.java
```diff
@@ -0,0 +1,243 @@
...
+                    case "max_raw_ttf":
+                        constantScore = ttf_stats.getMax();
+                        break;
+                    case "mean_raw_ttf":
+                        constantScore = ttf_stats.getMean();
+                        break;
+                    case "stddev_raw_ttf":
+                        constantScore = ttf_stats.getStdDev();
+                        break;
+
```
nomoa: could you add :
case "unique_terms_count": constantScore = terms.size(); break;

Thanks!

src/main/java/com/o19s/es/explore/ExplorerScorer.java
```diff
@@ -0,0 +1,100 @@
...
+
+        // Grab freq from subscorer, or the children if available
+        if(subScorer.getChildren().size() > 0) {
+            for(ChildScorer child : subScorer.getChildren()) {
+                if(child.child.docID() == docID()){
+                    tf_stats.add(subScorer.freq());
+                }
+            }
+        } else {
+            if(subScorer.docID() == docID()) {
```
nomoa: I think you can safely change this if to an assert:
`assert subScorer.docID() == docID();`

score is not supposed to be called if the iterator is not properly positioned.

PR #56, opened 2017-07-21
Add 2 new dec tree impl

nomoa:
This an attempt to optimize the naive implementation.
Results are not particularly great, the naive implementation remains
very fast. The prim implementation seems to perform slightly better
on server class CPUs and consumes 3 times less mem.
The binbin one is definetely not working well, it shows better perf
only on very large trees.


ebernhardson:
This is stalled because we aren't sure these implementations are actually that much better. They save some memory, but the evaluation latency is equal or slightly worse.

I have a WIP evaluator done with JNI at https://github.com/ebernhardson/elasticsearch-learning-to-rank/commits/branchless-trees that does manage to provide a 5x latency reduction in tests, but to get that speedup we need to be able to evaluate multiple documents in parallel(not thread parallel, but data parallel taking advantage of the out of order evaluation of modern x86 cpus). This will require a bit more upfront work for the rescorer to provide multiple documents at a time.

nomoa:
closing, will work on a custom rescore phase to allow bulk scoring so that Erik's work can be be integrated more easily.

PR #57, opened 2017-07-25
Don't use subWeights when scoring collection scoped stats

nomoa:
It prevents outputing the constant score for docs that do not match the nested
query.


nomoa:
@worleydl mind having a quick look?

src/test/java/com/o19s/es/explore/ExplorerQueryTests.java
```diff
@@ -135,7 +136,7 @@ public void testUniqueTerms() throws Exception {
         // Verify score is 4 (4 unique terms)
```
worleydl: Nitpicky, maybe update this comment since the score is changed. Also tq5 is no longer used, I included it before to double-check duplicate terms weren't bumping the unique term count.

nomoa: done

nomoa: oops sorry, I completely overlooked the comment on tq5...

PR #63, opened 2017-07-27
Elastic 5.5 support

nomoa:
- merge master
- add support for 5.5 (InternalSearchHit -> SearchHit)
- remove usage of deprecated native scripts


nomoa:
@worleydl @gehel mind having a quick look? Thanks!

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -50,6 +50,7 @@
     public static final ParseField STORE_NAME = new ParseField("store");
     public static final ParseField PARAMS = new ParseField("params");
 
+    private final transient FeatureStoreLoader storeLoder;
```
worleydl: Should this be storeLoader?

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -84,11 +93,12 @@ protected void doWriteTo(StreamOutput out) throws IOException {
...
 
-    public static Optional<StoredLtrQueryBuilder> fromXContent(QueryParseContext context) throws IOException {
+    public static Optional<StoredLtrQueryBuilder> fromXContent(FeatureStoreLoader storedLoader,
+                                                               QueryParseContext context) throws IOException {
         XContentParser parser = context.parser();
-        final StoredLtrQueryBuilder builder;
+        final StoredLtrQueryBuilder builder =  new StoredLtrQueryBuilder(storedLoader);
         try {
-            builder = PARSER.parse(context.parser(), null);
+            PARSER.parse(context.parser(), builder,null);
```
worleydl: Whitespace nitpick

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
nomoa: doh... you're right fixing.

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
nomoa: done

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -67,9 +68,17 @@
         declareStandardFields(PARSER);
     }
 
-    public StoredLtrQueryBuilder() {}
-    public StoredLtrQueryBuilder(StreamInput input) throws IOException {
+    public StoredLtrQueryBuilder() {
```
gehel: I assume that the default empty constructor is needed for some reason, but can't actually be used? And it must not throw an exception at construction time? A comment explaining why we allow building what looks like an unusable StoredLtrQueryBuilder would be welcomed.

nomoa: fair point

gehel: Is it really only for testing? If that's the case, wouldn't it make more sense to inject this fake store loader in the test? Or at least tag this constructor with `@VisibleForTesting`?

nomoa: I see what you mean, I'll move this behaviors to the test, it'll me give a chance to throw some hints on how to use it in integration tests.

PR #65, opened 2017-08-03
Precompiled Expression Features

worleydl:
Adds support for specifying expressions by setting the template language to "derived_expression".


src/main/java/com/o19s/es/ltr/query/FeatureVectorQuery.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.Weight;
+
+import java.io.IOException;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Supplier;
+
+public class FeatureVectorQuery extends Query {
```
nomoa: nitpick: maybe DerivedExpressionQuery would be more appropriate?

src/main/java/com/o19s/es/ltr/query/FeatureVectorQuery.java
```diff
@@ -0,0 +1,215 @@
...
+        if (!sameClassAs(obj)) {
+            return false;
+        }
+        FeatureVectorQuery that = (FeatureVectorQuery) obj;
+        return Objects.deepEquals(expression, that.expression);
+    }
+
+    @Override
+    public int hashCode() {
+    return 31 * classHash() + Objects.hash(expression);
```
nomoa: nitpick: indent

src/test/java/com/o19s/es/ltr/query/StoredLtrQueryIT.java
```diff
@@ -56,6 +59,8 @@ public void testFullUsecase() throws Exception {
                 QueryBuilders.matchQuery("field1", "{{query}}").toString()));
         addElement(new StoredFeature("feature2", Collections.singletonList("query"), "mustache",
                 QueryBuilders.matchQuery("field2", "{{query}}").toString()));
+        addElement(new StoredFeature("feature3", Collections.singletonList("query"), "derived_expression",
+                "(feature1 + feature2) * 5"));
```
nomoa: maybe you can test:
- set feature3 weight to 10
- use expression `signum(feature1 - feature2)` => should return -1 if feature2 if non zero, 1 if feature1 is non zero
Then you could assert that the score is > 10 if negativeScore == false, < -10 is negativeScore is true

src/main/java/com/o19s/es/ltr/query/FeatureVectorQuery.java
```diff
@@ -0,0 +1,215 @@
...
+                    return new FVDoubleValuesSource(supplier, features.featureOrdinal(name));
+                }
+            };
+
+            try {
+                DoubleValuesSource src = expression.getDoubleValuesSource(bindings);
+                DoubleValues values = src.getValues(context, null);
+                values.advanceExact(doc);
+                return Explanation.match((float) values.doubleValue(), "Evaluation of derived expression: " + expression.sourceText);
+            } catch (IOException ex) {
```
nomoa: feel free to update the signature on the base class and not catch IOException here, like you did for the scorer method

src/main/java/com/o19s/es/ltr/query/FeatureVectorQuery.java
worleydl: Done.

src/main/java/com/o19s/es/ltr/query/FeatureVectorQuery.java
worleydl: Fixed.

src/main/java/com/o19s/es/ltr/query/FeatureVectorQuery.java
worleydl: Done.

src/test/java/com/o19s/es/ltr/query/StoredLtrQueryIT.java
worleydl: Done.

worleydl:
Added a derived_expression feature to the LoggingIT tests and it seems to be working as expected.

nomoa:
Looks good to me, @gehel would you have few mins to take a quick look?
Thanks!

licenses/antlr4-runtime-4.5.1-1.jar.sha1
```diff
@@ -0,0 +1 @@
+66144204f9d6d7d3f3f775622c2dd7e9bd511d97
```
gehel: Why do we store the SHA1 of the licenses? It's probably obvious... but slow brain today...

src/main/java/com/o19s/es/ltr/feature/store/PrecompiledExpressionFeature.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        PrecompiledExpressionFeature that = (PrecompiledExpressionFeature) o;
+
+        if (!name.equals(that.name)) return false;
+        return !expression.equals(that.expression);
```
gehel: The negation looks strange here. 2 objects with the same expression are not equals?

Side note, using Objects.equals() is an easy way to write an equals method:

return Objects.equals(name, that.name)
    && Objects.equals(expression, that.expression);

src/main/java/com/o19s/es/ltr/feature/store/PrecompiledExpressionFeature.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        int result = name.hashCode();
```
gehel: Using Objects.hash() is a bit more compact and help not doing strange things with null references.

`return Objects.hash(name, expression)`

src/main/java/com/o19s/es/ltr/query/DerivedExpressionQuery.java
```diff
@@ -0,0 +1,207 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (!sameClassAs(obj)) {
+            return false;
+        }
+        DerivedExpressionQuery that = (DerivedExpressionQuery) obj;
```
gehel: Stupid question so that I can learn something, why are features not part of equality?

src/main/java/com/o19s/es/ltr/utils/Scripting.java
```diff
@@ -0,0 +1,83 @@
...
+        if (sm != null) {
+            sm.checkPermission(new SpecialPermission());
+        }
+        return AccessController.doPrivileged(new PrivilegedAction<Expression>() {
+            @Override
+            public Expression run() {
+                try {
+                    // snapshot our context here, we check on behalf of the expression
+                    AccessControlContext engineContext = AccessController.getContext();
+                    ClassLoader loader = getClass().getClassLoader();
```
gehel: It looks like getting the class loader should be extracted in its own method, and that `getClass().getClassLoader()` should be guarded by an `if (sm == null)`. Unless we do expect an exception on that call... in which case, we probably want a few comments explaining what is going on.

licenses/antlr4-runtime-4.5.1-1.jar.sha1
worleydl: Was just copying over how the lang-expression module handles the licenses.  No particular reason on my end.

src/main/java/com/o19s/es/ltr/feature/store/PrecompiledExpressionFeature.java
worleydl: Fixed.

src/main/java/com/o19s/es/ltr/feature/store/PrecompiledExpressionFeature.java
worleydl: Thanks for the tip, cleaned up.

src/main/java/com/o19s/es/ltr/query/DerivedExpressionQuery.java
worleydl: Not a stupid question, they should be there! Added.

src/main/java/com/o19s/es/ltr/query/DerivedExpressionQuery.java
```diff
@@ -0,0 +1,208 @@
...
+        return Objects.deepEquals(expression, that.expression)
+                && Objects.deepEquals(features, that.features);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(expression, features);
```
gehel: Note that you've dropped the `classHash()` from the hash. Looking at the code, this seems to be fine, but what do I know...

nomoa: right I missed that one, not a big deal I think because hashCode&equals should never be called at runtime (they're meant for caching but we never cache those)
Maybe worth fixing on a followup PR, it's probably cleaner.

gehel: I don't think the classHash() is actually needed. At lest, it looks like the usual equals()/hashCode() contract are OK without it.

PR #66, opened 2017-08-04
Ensure decision trees emit scores gt 0

ebernhardson:
Negative scores can cause interesting, unexpected, side effects in
elasticsearch. Add a constant value to all leafs to ensure we always
output a positive score.

### DIFF
--- a/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
+++ b/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
@@ -28,6 +28,7 @@ import java.util.Objects;
  * May be slow when the number of trees and tree complexity if high comparatively to the number of features.
  */
 public class NaiveAdditiveDecisionTree extends DenseLtrRanker implements Accountable {
+    static final float MIN_LEAF_VALUE = 0.1F;
     private static final long BASE_RAM_USED = RamUsageEstimator.shallowSizeOfInstance(Split.class);
 
     private final Node[] trees;
@@ -43,7 +44,7 @@ public class NaiveAdditiveDecisionTree extends DenseLtrRanker implements Account
      * @param weights the respective weights
      * @param modelSize the modelSize in number of feature used
      */
-    public NaiveAdditiveDecisionTree(Node[] trees, float[] weights, int modelSize) {
+    NaiveAdditiveDecisionTree(Node[] trees, float[] weights, int modelSize) {
         assert trees.length == weights.length;
         this.trees = trees;
         this.weights = weights;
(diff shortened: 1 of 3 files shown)


ebernhardson:
@nomoa could you take a look at this?

nomoa:
Negative scores are per se are not a bad thing, the problem is more on the context they are used.
This is because we assume that the ltr query will be in a rescore context where we completely inhibit first pass scores to 0.
I think this the issue, there's no easy way in elastic to inhibit the first pass score if we use multiple rescore  windows and one of them is greater than the last ltr one.
So while ensuring all scores to be at least > 0 we still have the problem of not being able to properly inhibit first pass scores.
In short I have no objection with this but I feel like we should provide a system where it's easy to rescore only the pages we want and still guarantee consistent ordering for pages outside the last rescore window.
Maybe a custom ltr rescore phase would do the the trick?

ebernhardson:
Negative scores are a problem even in a rescore query with a query_weight > 0 (i suppose it could be - though?). The issue would be that if you for example set the rescore query weight to 100, and the query weight to .1, the negative values would still come after the regular query values when they should probably come first.

src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
```diff
@@ -155,4 +166,63 @@ public long ramBytesUsed() {
...
+         *
+         * @param tree A tree to fix the minimum value of
+         * @return A new tree with fixed minimum value.
+         */
+        private Node fixMinLeafValue(Node tree) {
+            float minLeaf = minLeaf(tree);
+            if (minLeaf >= MIN_LEAF_VALUE) {
+                return tree;
+            } else {
+                return addToLeafs(tree, -minLeaf + MIN_LEAF_VALUE);
```
nomoa: if meanLeaf is [0,MIN_LEAF_VALUE] you'll have scores lower than MIN_LEAF_VALUE, you still ensure that they're all positive but not lower than MIN_LEAF_VALUE

ebernhardson:
Hmm, thinking about this some more i think you are right, this simply papers over the problem with a half solution while a custom rescore context should be able to solve the problem completely, without having to have documentation about specific limitations and work aronuds.

nomoa:
My bad I don't know why I thought I could provide a custom rescorer but this is not the case... This will be available in 6.1.
So now I'm not really sure on what we should do...
One way to solve this issue is to normalize the scores between bounds : 
- bound query score between [0, 1]
- bound ranker score between [1,2]
Then using a sum should ensure that the ranker score is kept for the top-N while remaining docs would keep their original ranking.

To normalize we could use the [logistic function](https://en.wikipedia.org/wiki/Logistic_function) which can normalize an arbitrary value between bounds while ensuring same ranking across replicas. Or even the saturation function can be tweaked to do this as well.

To do this right now we could write a custom lucene query that wraps and normalizes an arbitrary lucene query but that means we will pay the normalization price on all retrieved docs, while in theory we only need to normalize the top docs.
I'm pretty sure we can do this as a function score with script score as well.

I'm undecided on what we should do now... wait for 6.1 or go ahead and pay the price of normalizing all retrieved docs...

softwaredoug:
Closing for now, revisit for 6.1. We should document the paging weirdness.

Issue #35, opened 2017-05-03
Model Language: Change from ordinals to names


sstults:
The names may refer to search templates.

epugh:
@worleydl can you investigate if this is done?

nomoa:
This is I think the last bits that still need to be done for v1, currently few model parsers (linear, xgboost) can understand named features. The ranklib model languages (AFAIU) currently only supports ordinals.
Maybe we could rename as: support named features in ranklib model formats?

softwaredoug:
Largely done, however we will need to support ordinal-based ranklib forever I suspect. So I think we will need to fix issue #85

Issue #15, opened 2017-03-08
Kill the Ranklib dependency


softwaredoug:
Related to #14. Scientist say we only use 1% of our Brains err I mean Ranklib.

So when you focus on what actually matters, we don't really care about most of the models. We don't really care about the training side of Ranklib. And when you take all that stuff away you get a very small set of code for evaluating some models. 

This issue is to kill the dependency and just directly include the models/code we care about

dacox:
seems like the milestone is a little out of date since `v1.1` is already out

softwaredoug:
Good call @dacox... I'm going to close this issue as I don't think its a huge priority

PR #74, opened 2017-08-31
Ranklib Changes

worleydl:
Addresses issues #15 and #35 

Dependency on ranklib removed, needed code has been brought into the project with unnecessary items removed.

LegacyRanklibModelParser: Parses models using ordinals
RanklibModelParser: Parses using names


src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
```diff
@@ -20,6 +20,7 @@
 import com.o19s.es.ltr.ranker.LtrRanker;
 import com.o19s.es.ltr.ranker.parser.LtrRankerParser;
 import com.o19s.es.ltr.ranker.parser.LtrRankerParserFactory;
+import com.o19s.es.ltr.ranker.ranklib.learning.FEATURE_TYPE;
```
nomoa: looks like it's unused

src/main/java/com/o19s/es/ltr/ranker/ranklib/learning/CoorAscent.java
```diff
@@ -0,0 +1,192 @@
+/*
+ * Copyright [2017] Wikimedia Foundation
```
nomoa: maybe we should remove this copyright notice?

worleydl:
In regards to licensing, ranklib is BSD which supports being re-licensed under Apache, the original copyright needs to be included in the source to satisfy BSD.  I think the current headers are fine given we modified the code a bit (Named featured support and slight changes to make the code compatible with the gradle build).

nomoa:
@worleydl thanks for the info!
This PR looks good to me, I haven't run the build myself to make sure it passes but this should be good to merge.
note that I haven't reviewed the python code carefully.
Side note: there is a comment on docs/models.json stating that ranklib is deprecated maybe we can remove that line now and add a quick note on the new type `ranklib+names`?

Thanks!

softwaredoug:
I haven't looked at this, but one note: we found a lot of model parsing bugs with Ranklib and fixed them in RankyMcRankFace, which became the dependency we used for Elasticsearch Learning to Rank. Have any of those been incorporated here?

For example, straight Ranklib was taking ridiculous amounts of time to parse a large random forest ranker, and we reduced the time to a few seconds.

src/main/java/com/o19s/es/ltr/ranker/ranklib/learning/CoorAscent.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * @author vdang
+ * 
+ * This class implements the linear ranking model known as Coordinate Ascent. It was proposed in this paper:
+ *  D. Metzler and W.B. Croft. Linear feature-based models for information retrieval. Information Retrieval, 10(3): 257-274, 2007.
+ */
+public class CoorAscent extends Ranker {
```
ebernhardson: This should be unnecessary, at ranking time coordinate ascent is just a weighted sum like the current linear model

src/main/java/com/o19s/es/ltr/ranker/ranklib/learning/LinearRegRank.java
```diff
@@ -0,0 +1,146 @@
...
+    public LinearRegRank(List<RankList> samples, int[] features, MetricScorer scorer)
+    {
+        super(samples, features, scorer);
+    }
+    public void init()
+    {
+        PRINTLN("Initializing... [Done]");
+    }
+
+    public double eval(DataPoint p)
```
ebernhardson: this is also a simple weighted sum, like the pre-existing linear model

softwaredoug:
I would like to postpone this PR till after releasing 1.0 as it's rather large, and right now I think we can get by with the ranklib dependency. Any opposition?

worleydl:
No opposition from me.  Sorry for taking a bit to get things wrapped up, I
should have free time next week.

softwaredoug:
Delaying to 1.1...

Issue #83, opened 2017-09-09
Feature/Feature Set relationship confused somewhat in API


softwaredoug:
Initially when I developed the demo, I had assumed from the structure of this request:

```
{
  "name": "my_feature_set",
  "features" : [
    {
      "name": "my_feature",
      "params": ["query_string"],
      "template_language": "mustache",
      "template" : {
        "match": {
          "field": "{{query_string}}"
        }
      }
    }
  ]
}
```

That features were "owned by" a feature set. IE a strong composition relationship. This is further reinforced by requiring feature sets to be created with new features (see #82). For example, in this sort of relationship I assumed if you delete a feature set, you would also delete all associated features. Or you would do `GET _ltr/my_feature_set/feature/my_feature`.

However, it appears on further study that features exist independently of feature sets, and the relationship is more associative (which makes a lot of sense to me).

Can we change this API to avoid the confusion by creating a clearer associative relationship? For example, I would suggest something that did not automatically create features in the set, such as notionally something like:

```

PUT _ltr/_feature/my_feature
   {
      "name": "my_feature",
      "params": ["query_string"],
      "template_language": "mustache",
      "template" : {
        "match": {
          "field": "{{query_string}}"
        }
      }
   }

PUT _ltr/_featureset
{
  "name": "my_feature_set",
  "features" : [{
      "name": "my_feature"
   }]
}
```

I could live with the current way things are done, but as I'm documenting I'm seeing how it can be confusing.

At the very least, I think it would be good to be able to support a feature set creation syntax like what I've proposed above where feature sets are created from existing features. (maybe one exists and I missed it?).

Thoughts? Am I missing something?

softwaredoug:
I think perhaps the relationship is best termed as "copying to" rather than associative. So it is a composition relationship, but features are copied into feature sets, and then those features are copied into models when models are created. Is that correct?

softwaredoug:
Perhaps the best solution would be to remove the feature API and allow feature creation only in the context of feature sets?

- Create a feature set?
- Append to a feature set with a body?

nomoa:
Agreed this is confusing. I think we should document a normal usecase starting from featuresets (ignoring the _feature endpoint).
Storing individual feature can be removed (no objections on my side). But we could see the `_feature` endpoint as a way to store popular features that are frequently added to featuresets.
While we discuss this I'll go ahead a implement a way to append features to a set.

softwaredoug:
[...]

PR #87, opened 2017-09-15
Update featureset with features provided in request body

nomoa:
Allows to update featureset by adding features in the request body
rather than providing a prefix query.
Initially it was only possible to update the featureset by adding
features from the store.
It's now possible to provide the feature definitions directly in
the request body bypassing the feature store.
Added an option to merge the features, it permits to replace a
feature instead of throwing an error.

relates #83

### DIFF
--- a/src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureSet.java
+++ b/src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureSet.java
@@ -175,6 +175,36 @@ public class StoredFeatureSet implements FeatureSet, Accountable, StorableElemen
         return new StoredFeatureSet(name, newFeatures);
     }
 
+    /**
+     * Generate a new set by merging features with this set, feature with same names are updated new features are appended.
+     * Ordinals of existing features are kept.
+     *
+     * @param mergedFeatures the list of features to merge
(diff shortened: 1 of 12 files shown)


src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureSet.java
```diff
@@ -175,6 +175,36 @@ public StoredFeatureSet append(List<StoredFeature> features) {
         return new StoredFeatureSet(name, newFeatures);
     }
 
+    /**
+     * Generate a new set by merging features with this set, feature with same names are updated new features are appended.
```
softwaredoug: Hmm I thought we issued errors when we tried to append a feature to a featureset when a name was already taken.

nomoa: This is on purpose, you'll have the choice to merge the features.
By default it'll be like you said, features are always appended and will fail if a name already exists.
But adding `?merge=true` to the request url will instruct the plugin to merge and overwrite features.

softwaredoug: Oh I see. Understood

PR #88, opened 2017-09-16
Fixes #85 -- logs features as a list

### LINKED ISSUE #85: Logging response include ordinal
As discussed with @nomoa, plain-old RankLib only thinks in ordinals. Should we change the log response to optionally include the ordinal of the feature?

softwaredoug:
Change the dictionary logging structure to a list. This ensures you can use the ordinal when logging for tools like ranklib that want that info. 

Example:

```
      "fields": {
               "_ltrlog": [
                  {
                     "log_entry1": [
                        {
                           "name": "body_query",
                           "value": 9.8376875
                        },
                        {
                           "name": "title_query",
                           "value": 12.318446
                        }
                     ]
                  }
               ]
```

When no value present (and `missing_as_zero` is `false`) then feature is logged with "value" key missing, ie:

```
      "fields": {
               "_ltrlog": [
                  {
                     "log_entry1": [
                        {
                           "name": "body_query"
                        },
                        {
                           "name": "title_query",
                           "value": 12.318446
                        }
                     ]
                  }
               ]
```

### DIFF
--- a/src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
+++ b/src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
@@ -215,9 +215,6 @@ public class LoggingFetchSubPhase implements FetchSubPhase {
 
         @Override
         public void accept(int featureOrdinal, float score) {
-            if (currentLog.isEmpty()) {
-                rebuild();
-            }
             currentLog.get(featureOrdinal).put("value", score);
         }
 
--- a/src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
+++ b/src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
@@ -179,24 +179,46 @@ public class LoggingFetchSubPhase implements FetchSubPhase {
         private static final String FIELD_NAME = "_ltrlog";
         private final String name;
         private final FeatureSet set;
-        private final Map<String, Float> initialLog;
-        private Map<String, Float> currentLog;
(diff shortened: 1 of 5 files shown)


src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
```diff
@@ -179,24 +179,46 @@ void doLog(BooleanQuery query, List<HitLogConsumer> loggers, IndexSearcher searc
...
+                ini.add(i, defaultKeyVal);
             }
-            initialLog = Collections.unmodifiableMap(ini);
+            currentLog = ini;
         }
 
         @Override
         public void accept(int featureOrdinal, float score) {
-            currentLog.put(set.feature(featureOrdinal).name(), score);
+            if (currentLog.isEmpty()) {
```
nomoa: nitpick: this if should not be needed and can be removed.
currentLog cannot be empty (initialized at null and then rebuild is always called before collecting)

src/test/java/com/o19s/es/ltr/logging/LoggingFetchSubPhaseTests.java
```diff
@@ -41,6 +41,7 @@
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.util.LuceneTestCase;
 import org.apache.lucene.util.TestUtil;
+import org.elasticsearch.common.collect.Tuple;
```
nomoa: nitpick: uneeded

softwaredoug:
@nomoa removed unneeded code

PR #89, opened 2017-09-19
Add support for feature validation

### LINKED ISSUE #84: To avoid costly errors, validate features when they're added
If you have valid JSON, but a semantically invalid query (for example it's easy to add "query") like below when the plugin expects things at the "match" level:

```
 {
      "name": "user_rating",
      "type": "feature",
      "feature": {
         "name": "user_rating",
         "params": [],
         "template": {
            "query": {
               "match": {
                  "title": "{{keywords}}"
               }
            }
         }
      }
   }
```

Say this is feature "1" and you've built up a large feature set of 100 queries. Then you go to execute the feature set. You'll get a bug that the sltr query could not be executed.

As feature sets are append-only your whole feature set is screwed up and you have to start over. Is there any way to validate this is a valid query (ie attempt to parse it) before letting the feature be cerated

nomoa:
Introduce a new json block to ask the plugin to fully validate features
prior to storing elements.
This still requires the user to explicitly set the "validation" which
contains the information needed to validate features:
- the index to run the query on (for mapping validation in some cases)
- the template param values

NOTE: that I had to change the body format to include the new "validation".
See changes in the docs folder for more details.

It might be interesting (1.1?) to add a new specific rest endpoint to validate
stored features/set/models.

closes #84

### DIFF
--- a/src/main/java/com/o19s/es/ltr/action/TransportFeatureStoreAction.java
+++ b/src/main/java/com/o19s/es/ltr/action/TransportFeatureStoreAction.java
@@ -24,16 +24,22 @@ import com.o19s.es.ltr.feature.store.StoredFeature;
 import com.o19s.es.ltr.feature.store.StoredFeatureSet;
 import com.o19s.es.ltr.feature.store.StoredLtrModel;
 import com.o19s.es.ltr.feature.store.index.IndexFeatureStore;
+import com.o19s.es.ltr.query.ValidatingLtrQueryBuilder;
 import com.o19s.es.ltr.ranker.parser.LtrRankerParserFactory;
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.index.IndexAction;
 import org.elasticsearch.action.index.IndexRequest;
+import org.elasticsearch.action.search.SearchAction;
+import org.elasticsearch.action.search.SearchRequestBuilder;
+import org.elasticsearch.action.search.SearchResponse;
+import org.elasticsearch.action.search.ShardSearchFailure;
 import org.elasticsearch.action.support.ActionFilters;
 import org.elasticsearch.action.support.HandledTransportAction;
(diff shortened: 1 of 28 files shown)


src/main/java/com/o19s/es/ltr/action/TransportFeatureStoreAction.java
```diff
@@ -121,7 +113,7 @@ private IndexRequest buildIndexRequest(Task parentTask, FeatureStoreRequest requ
         return indexRequest;
     }
 
-    private void validate(FeatureStoreRequest request) {
+    private void validateAndSave(FeatureStoreRequest request, Task task, ActionListener<FeatureStoreResponse> listener) {
```
gehel: Looking at the code, this seems to be `validateOrSave()`. I think this would all be more readable by splitting in 2 methods (`validate` / `save`) and move the `if (request.getValidation() != null` to `doExecute()`. But I might be missing some dependency...

src/main/java/com/o19s/es/ltr/feature/FeatureValidation.java
```diff
@@ -0,0 +1,105 @@
...
+
+        FeatureValidation that = (FeatureValidation) o;
+
+        if (!index.equals(that.index)) return false;
+        return params.equals(that.params);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = index.hashCode();
```
gehel: Minor style issue, feel free to ignore.

I find that [`Objects.hash()`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-) and [`Objects.equals()`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-) make for nicer implementations. Note that IntelliJ has a template to generate `equals()` / `hashCode()` using the Objects helper methods.

PR #110, opened 2017-11-13
Fixes #104 -- Ranklib models use feature set size when available

### LINKED ISSUE #104: Error in DenseDataPoint::setFeatureValue(): feature (id=95) not found.
I am getting this error when trying to load a LambdaMART model that does not use all features in my featureset. Describing in more details, I did the following steps:

1. Created a featureset of 371 features
2. Trained a LambdaMART model on Ranklib. This model uses 133 features (not all features of my feature set)
3. Loaded the model (output file generated by Ranklib)
4. Used the model in the sltr query and as a return I got the error `Error in DenseDataPoint::setFeatureValue(): feature (id=95) not found.`

softwaredoug:
Create a feature set with multiple features. When a Ranklib Model with a single feature is evaluated, such as:

```
                              <ensemble>
                                <tree id="1" weight="0.1">
                                  <split>
                                    <feature> 1 </feature>
                                    <threshold> 1.0 </threshold>
                                    <split pos="left">
                                      <output> -2.0 </output>
                                    </split>
                                    <split pos="right">
                                     <output> 2.0 </output>
                                    </split>
                                  </split>
                                </tree>
                              </ensemble>
```

The Ranklib Ranker will ask the model for the number of features (in this case 1) when creating a dense feature vector. But it will attempt to evaluate all features in the feature set (say 4). When features 2...4 are evaluated, out of bounds error occurs in DenseProgramatticDataPoint

This PR uses the feature set's feature size when creating vectors for Ranklib.

I have yet to recreate this with xgboost and others, but it's an area worth investigating


src/test/resources/rest-api-spec/test/fstore/80_search.yaml
```diff
@@ -0,0 +1,120 @@
+---
+setup:
```
nomoa: nitpick: the test name could be bit more explicit, e.g. setup simple test models with a test indice

nomoa:
I've just rebased your change, will merge as soon as CI is happy.

PR #114, opened 2017-12-04
Fixes #113 - Fix documentation and add handler on POST /_ltr/{store}

### LINKED ISSUE #113: Problem creating feature: contains unrecognised parameter store
I am trying to play with the plugin, however in step two, creating a feature set I get an error. I used the following command:

```
POST _ltr/_featureset
{
  "validation": {
     "params": {
         "keywords": "java"
     },
     "index": "blog"
    },
   "featureset": {
        "name": "better_blogs",
        "features": [
            {
                "name": "title_query",
                "params": [
                    "keywords"
                ],
                "template": {
                    "match": {
                        "title": "{{keywords}}"
                    }
                }
            }
        ]
   }
}
```

I get this response:

```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "request [/_ltr/_featureset] contains unrecognized parameter: [store]"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "request [/_ltr/_featureset] contains unrecognized parameter: [store]"
  },
  "status": 400
}
```
I verified the plugin is installed, using the same version as the cluster (5.6.4). Any help on trying to overcome this problem would be nice. Not sure if I am doing something wrong or not. Following a long with the documentation.

Thanks

nomoa:
closes #113

### DIFF
--- a/docs/building-features.rst
+++ b/docs/building-features.rst
@@ -110,7 +110,7 @@ Create a feature set
 You can create a feature set simply by using a POST. To create it, you give a feature set a name and optionally a list of features::
 
 
-    POST _ltr/_featureset
+    POST _ltr/_featureset/more_movie_features
     {
        "featureset": {
             "name": "more_movie_features",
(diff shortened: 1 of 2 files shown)


ebernhardson:
Perhaps worth also thinking about the additional question posed in 113, if the name of the featureset is in the url does it also need to be in the request body? It seems it can't hurt to keep it, but it should be optional (not sure if it already is, but if it is optional the docs should leave it out)

nomoa:
I can try and see how elastic reacts but I added name to make sure it does not conflict with other endpoints like creating a feature: `POST /_ltr/store_name`

ebernhardson:
I think it's ok to keep the name in the url, i was thinking to remove the name from the document body. I suppose it still has to be there when writing the document out to the feature store index, but it seems the code could inject it?

nomoa:
Sure, I'll have a look.

PR #115, opened 2017-12-05
Make element name optional

nomoa:
Inject the name found in the url part.

relates #113


ebernhardson:
This ended up being a bigger change than I expected, sorry about that! Looks good though and should make things a little more obvious without the duplication.

PR #120, opened 2017-12-18
Fixes #118 properly handle cache state on store and model deletion

### LINKED ISSUE #118: Model cache not cleared, stale models executed
In my demo, I'm deleting and rebuilding _ltr, the feature set, and model regularly. I notice that even though I inspect the model and it only contains teh 3 features I expect, running the model shows 5 features when `explanation: true`. Manually clearing the feature cache fixes this.

- Does deleting `_ltr` clear the cache? Deleting the model? Overwriting the model?

nomoa:
The cache was only properly maintained on UPDATE actions.
Add proper cache invalidation on:
- store deletion
- model/set/feature deletion

We still do not maintain the cache properly if the user attempts to change
the store content directly bypassing our rest endpoints (creating model
using the elasticsearch index API).

Fixes #118


nomoa:
@softwaredoug this should be good now.
Sorry for being late on this one!

PR #125, opened 2018-01-31
[WIP] X-Pack support

### LINKED ISSUE #116: security_exception: [ltr:featurestore/data] is unauthorized for user
I am trying out your plugin for the first time and going through the demo. I can get the `tmdb` index fully populated but am finding problems when it comes to interacting with the plugin's endpoints.  I get a `security_exception` on most endpoints.

```
demo$ python loadFeatures.py 
DELETE http://elastic:changeme@localhost:9200/_ltr
200
PUT http://elastic:changeme@localhost:9200/_ltr
200
POST http://elastic:changeme@localhost:9200/_ltr/_featureset/movie_features
{
  "featureset": {
    "name": "movie_features",
    "features": [
      {
        "name": "1",
        ..snip!..
      }
    ]
  }
}
403
{"error":{"root_cause":[{"type":"security_exception","reason":"action [ltr:featurestore/data] is unauthorized for user [elastic]"}],"type":"security_exception","reason":"action [ltr:featurestore/data] is unauthorized for user [elastic]"},"status":403}
```

I am running on [the base Docker image](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/docker.html) of Elasticsearch version 5.6.4.  The plugin version is `ltr-1.0.0-RC2-es5.6.4.zip`.  I'm running as a user with the `superuser` role as verified by:
```
demo$ curl 'elastic:changeme@localhost:9200/_xpack/security/user/elastic?pretty'
{
  "elastic" : {
    "username" : "elastic",
    "roles" : [
      "superuser"
    ],
    "full_name" : null,
    "email" : null,
    "metadata" : {
      "_reserved" : true
    },
    "enabled" : true
  }
}
```

Oddly enough, some plugin endpoints succeed, such as `DELETE /_ltr` and `PUT /_ltr` and `GET /_ltr/_featureset`.  I also tried both [old and new versions](https://github.com/o19s/elasticsearch-learning-to-rank/commit/67ee90893f819b0ec653d0998753c574a4b65d50) of feature set creation shown in the docs, namely `POST /_ltr/_featureset` and `POST /_ltr/_featureset/movie_features`.

Am I missing some initial set-up around permissions or user accounts?

ebernhardson:
* X Pack only allows our custom transport actions if we name them such
 that they live inside the cluster management namespace. Rename them
 all.
* When renaming cluster actions there doesn't seem to be a great way
 to handle BC. During a rolling restart to deploy this patch both
 cluster actions will be possible, but nodes will only support one
 or the other.
* Adjusted demo to support xpack. A support script creates
 appropriate admin and user roles then creates an ltr admin user. The
 rest of the scripts are adjusted to consistently use the host and
 username from settings.cfg.

Fixes #116

### DIFF
--- a/demo/prepare_xpack.py
+++ b/demo/prepare_xpack.py
@@ -1,0 +1,46 @@
+from elasticsearch_xpack import XPackClient
+import getpass
+import sys
+from utils import Elasticsearch
+
+if __name__ == "__main__":
+    if len(sys.argv) == 2:
+        password = getpass.getpass()
+    elif len(sys.argv) == 3:
+        password = sys.argv[2]
+    else:
+        sys.exit(-1)
+
+    username = sys.argv[1]
+
+    es = Elasticsearch(http_auth=(username, password))
+    xpack = XPackClient(es)
+
+    print("Configure ltr_admin role:")
+    res = xpack.security.put_role('ltr_admin', {
+        "cluster": ["all"],
+        "indices": [ {
+            "names": [".ltrstore*"],
+            "privileges": ["all"]
+        } ]
+    })
+    print(res)
+
+    print("Configure tmdb role:")
+    res = xpack.security.put_role('tmdb', {
+        'indices': [ {
+            "names": ["tmdb"],
+            "privileges": ["all"]
+        } ]
+    })
+    print(res)
+
+    print("Configure ltr_demo user:")
+    res = xpack.security.put_user('ltr_demo', {
+        'password': 'changeme',
+        'roles': ['ltr_admin', "tmdb"]
+    })
+    print(res)
+
+    print("\nRoles and user created. Be sure to update settings.cfg")
+
--- a/demo/settings.cfg
+++ b/demo/settings.cfg
@@ -1,2 +1,4 @@
 [DEFAULT]
 ESHost = http://localhost:9200
+ESUser = ltr_demo
+ESPassword = changeme
--- a/src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
+++ b/src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
@@ -171,11 +171,17 @@ public class LtrQueryParserPlugin extends Plugin implements SearchPlugin, Script
     public List<ActionHandler<? extends ActionRequest, ? extends ActionResponse>> getActions() {
         return unmodifiableList(asList(
                 new ActionHandler<>(FeatureStoreAction.INSTANCE, TransportFeatureStoreAction.class),
+                new ActionHandler<>(FeatureStoreAction.BC_INSTANCE, TransportFeatureStoreAction.BC.class),
                 new ActionHandler<>(CachesStatsAction.INSTANCE, TransportCacheStatsAction.class),
+                new ActionHandler<>(CachesStatsAction.BC_INSTANCE, TransportCacheStatsAction.BC.class),
                 new ActionHandler<>(ClearCachesAction.INSTANCE, TransportClearCachesAction.class),
+                new ActionHandler<>(ClearCachesAction.BC_INSTANCE, TransportClearCachesAction.BC.class),
                 new ActionHandler<>(AddFeaturesToSetAction.INSTANCE, TransportAddFeatureToSetAction.class),
+                new ActionHandler<>(AddFeaturesToSetAction.BC_INSTANCE, TransportAddFeatureToSetAction.BC.class),
                 new ActionHandler<>(CreateModelFromSetAction.INSTANCE, TransportCreateModelFromSetAction.class),
-                new ActionHandler<>(ListStoresAction.INSTANCE, TransportListStoresAction.class)));
+                new ActionHandler<>(CreateModelFromSetAction.BC_INSTANCE, TransportCreateModelFromSetAction.BC.class),
+                new ActionHandler<>(ListStoresAction.INSTANCE, TransportListStoresAction.class),
+                new ActionHandler<>(ListStoresAction.BC_INSTANCE, TransportListStoresAction.BC.class)));
     }
 
     @Override
--- a/src/main/java/com/o19s/es/ltr/action/ListStoresAction.java
+++ b/src/main/java/com/o19s/es/ltr/action/ListStoresAction.java
@@ -38,7 +38,7 @@ import java.util.stream.Collectors;
 
 public class ListStoresAction extends Action<ListStoresAction.ListStoresActionRequest,
         ListStoresAction.ListStoresActionResponse,ListStoresAction.ListStoresActionRequestBuilder> {
-    public static final String NAME = "ltr:featurestore/list";
+    public static final String NAME = "cluster:admin/ltr/featurestore/list";
     public static final ListStoresAction INSTANCE = new ListStoresAction();
 
     private ListStoresAction() {
--- a/demo/README.md
+++ b/demo/README.md
@@ -8,6 +8,18 @@ This demo requires
 
 - Python 3+
 - Python `elasticsearch` and `requests` libraries
+- Python `elasticsearch_xpack` libraries if xpack support is necessary
+
+## An aside: X Pack
+
+Using the LTR plugin with xpack requires configuring appropriate roles. These
+can be setup automatically by `prepare_xpack.py` which takes a username and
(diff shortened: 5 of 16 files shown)


ebernhardson:
This seems to work in my testing, but it's been pretty limited. The question about BC for cluster actions remains open. I could work out something to make the upgrade completely clean, but not sure it's worth the effort. Essentially we first have to release a version of the plugin that supports both action names and creates actions with the old names, then we have to do a second release that starts using the new names. Then at some point in the future the BC action names can be removed.

nomoa:
lgtm on my side, I'm fine removing BC support aswell since it's already partial and that none of the actions are required at runtime (by the sltr query I mean).
Basically the limitation will be: don't manage the ltr store while upgrading to this new version.

nomoa:
Remaining question is for some actions that are run from transport actions: https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/ltr/action/TransportFeatureStoreAction.java#L191

This relates to the suggestion made on discuss at https://discuss.elastic.co/t/plugin-development-custom-transport-actions-and-x-pack-integration/110805/6 .
Basically a quick way to check would be to see if there are any errors in the logs related to cache invalidation when overriding a featureset. I'm hoping that the Client injected in the transport actions is already granted with super user perms and it should be just fine.

src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
```diff
@@ -171,11 +171,17 @@ public ScriptEngine getScriptEngine(Settings settings, Collection<ScriptContext<
     public List<ActionHandler<? extends ActionRequest, ? extends ActionResponse>> getActions() {
         return unmodifiableList(asList(
                 new ActionHandler<>(FeatureStoreAction.INSTANCE, TransportFeatureStoreAction.class),
+                new ActionHandler<>(FeatureStoreAction.BC_INSTANCE, TransportFeatureStoreAction.BC.class),
```
nomoa: I don't see the corresponding code.
But I'm fine dropping any kind of BC support it since it'll cover only the usecase where an old node communicate to a new one.

purbon: I am not sure we should worry for now about breaking compatibility, however we should clearly document that. or even probably backport as update for other versions of elasticsearch.

src/main/java/com/o19s/es/ltr/action/ListStoresAction.java
```diff
@@ -38,7 +38,7 @@
 
 public class ListStoresAction extends Action<ListStoresAction.ListStoresActionRequest,
         ListStoresAction.ListStoresActionResponse,ListStoresAction.ListStoresActionRequestBuilder> {
-    public static final String NAME = "ltr:featurestore/list";
+    public static final String NAME = "cluster:admin/ltr/featurestore/list";
```
purbon: should this be cluster:admin? would this not be too much open privileges? what do you think?

demo/prepare_xpack.py
```diff
@@ -0,0 +1,46 @@
...
+
+    es = Elasticsearch(http_auth=(username, password))
+    xpack = XPackClient(es)
+
+    print("Configure ltr_admin role:")
+    res = xpack.security.put_role('ltr_admin', {
+        "cluster": ["all"],
+        "indices": [ {
+            "names": [".ltrstore*"],
+            "privileges": ["all"]
```
purbon: I wonder if we can provide probably more find grain privileges? not sure we need all, what do you think about something like: manage, monitor, read and write?

demo/prepare_xpack.py
```diff
@@ -0,0 +1,46 @@
+from elasticsearch_xpack import XPackClient
```
purbon: should we also have this script as part of some bin directory, some utility to setup the plugin with xpack?

purbon:
I found https://github.com/elastic/elasticsearch/blob/99f88f15c5febbca2d13b5b5fda27b844153bf1a/server/src/test/java/org/elasticsearch/transport/ActionNamesIT.java#L28 as interesting source of action names, you might probably already know 😄

src/main/java/com/o19s/es/ltr/action/ListStoresAction.java
ebernhardson: This would almost be best as something like `indices:data/read/ltr/featurestore/list`, as all this really does is read from the index. I'll have to poke around to see if we can actually use that though.

src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
ebernhardson: I also think that the BC support is too complex to be worthwhile. As david noted this kind of BC support would only allow old node to talk to new one, new talking to old would still fail. Getting around that means needing an intermediate version and 2 cluster restarts.  All for actions that are only administrative and not part of the core data collection/scoring operations.  I'm going to drop this code and add to the documentation somewhere.

demo/settings.cfg
```diff
@@ -1,2 +1,4 @@
 [DEFAULT]
 ESHost = http://localhost:9200
+ESUser = ltr_demo
+ESPassword = changeme
```
ebernhardson: These should be commented out by default

demo/prepare_xpack.py
ebernhardson: I didn't here as it was just the demo, but if we want to provide roles that we suggest to use on production clusters then certainly "all" is a bit too much. Also we should probably have separate `ltr_admin` and `ltr_query` roles, where `ltr_query` only needs read access to the index for when it needs to fetch a model not in cache. 

Its not particularly complicated so it probably makes sense to put the full featured setup here in the demo. Or even outside the demo and referenced from here? Not sure.

demo/prepare_xpack.py
```diff
@@ -0,0 +1,46 @@
...
+import sys
+from utils import Elasticsearch
+
+if __name__ == "__main__":
+    if len(sys.argv) == 2:
+        password = getpass.getpass()
+    elif len(sys.argv) == 3:
+        password = sys.argv[2]
+    else:
+        sys.exit(-1)
```
ebernhardson: Add usage as top level doc string, and print(__doc__) here.

ebernhardson:
I actually hadn't seen that list of action names, quite useful!

purbon:
@ebernhardson I tested the PR manually with several cases, and without the BC thing 😄 , it works quite good for me and makes sense to me as an initial solution path.

In my mind missing points for this PR to be more close to be merged are:

* have more find grand definitions for the action names, more on the idea of what we have commented in https://github.com/o19s/elasticsearch-learning-to-rank/pull/125#discussion_r165143002.

* document necessary setup to make this project work within the xpack land. Including how to upgrade from previous setup and versions.

bonus points for me are:

* Add reliable integration test with with security in mind.


what do you think?

src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java
purbon: good, I can also help with documenting and adding test to this tomorrow.

demo/prepare_xpack.py
purbon: I would probably add this as reference in the documentation.

src/main/java/com/o19s/es/ltr/action/ListStoresAction.java
purbon: I like that one, I would definitely revisit the names to make them less general, specially because xpack (to the top of my head) uses this names to decide of authorisation.

purbon:
@ebernhardson might i collaborate with you moving this forward? I would really like to merge it soon.

ebernhardson:
Yes, i'm not sure how it works on github  but if you want to move the patch forward feel free to update and propose a new version. I may or may not have time this weekend.

purbon:
@ebernhardson I would really like to help merging this soon, I will push a version in the next minutes with what I think should be minim stuff to be merged, we can always improve it later.

purbon:
Trying to help moving this forward I created #126 , let me know what do you think.

ebernhardson:
Closing in favor of purban's  continuation of this patch.

PR #133, opened 2018-02-12
Avoid NaN in query explorer stddev

ebernhardson:
The stddev calculation performs a division by zero when calculating
the stddev of a single element. This is technically correct for the
approximate definition of stddev, but not particularly useful for us.
Replacing the stddev with 0.0 in this case seems defensible, as we
are not approximating a domain, we know the full set of values. And
for this full set of values there is no deviation.


purbon:
@ebernhardson I see the usefulness of this change, however I am wondering if it would be interesting to have this done through a config option? returning zero in such a case, and not undefined is super reasonable, I am just wondering if people might be willing to activate this explicitly?

ebernhardson:
I think NaN is an incorrect result for stddev in this case. Per [Rapid calculation methods](https://en.wikipedia.org/wiki/Standard_deviation#Rapid_calculation_methods) The calculation being used is for calculating sample stddev, but we know in this case that we are not calculating stddev from a sample of the population, we have the entire population. Per the linked page maybe the algorithm used here should be changed from (n-1) to n?

nomoa:
Agreed, reading https://stats.stackexchange.com/questions/17890/what-is-the-difference-between-n-and-n-1-in-calculating-population-variance we have the entire population everytime so n-1 makes no sense apparently. We should go with /N.

purbon:
Makes sense to me!

El dt., 13 de febr., 2018 10.39, David Causse <notifications@github.com> va
escriure:

PR #137, opened 2018-02-16
Fail when logging NaN values

### LINKED ISSUE #136: Fail when trying to log NaN values
Happened because of #133 and the error was detected only during training.
It may be rare and only occurs with bogus queries but we should warn the user as soon as possible that something is not working as expected.

Producing NaN may lead to feature vectors containing `missing=false`, `value=NaN` which is very confusing.

nomoa:
It can be extremely misleading to allow logging NaN as it creates
a log entry missing:false, value:NaN.
We should detect this behavior as soon as possible to warn the user
that their query is broken.

fixes #136

### DIFF
--- a/src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
+++ b/src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
@@ -48,14 +48,14 @@ import java.util.Optional;
 
 public class LoggingFetchSubPhase implements FetchSubPhase {
     @Override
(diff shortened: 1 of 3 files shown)


softwaredoug:
Is this PR still useful?

nomoa:
@softwaredoug yes I believe so, it's certainly not a critical feature but might help in some corner cases to early warn the user that something is wrong in their feature set.

purbon:
I agree with such assessment.

/ Pere

El dg., 15 d’abr., 2018 22.00, David Causse <notifications@github.com> va
escriure:

PR #143, opened 2018-03-10
ES 6.2.2

softwaredoug:
TODO

- [x] Figure out how to deal with `tf` in ExplorerScorer
- [x] Review whether isCacheable is correct for each weight we implement

### DIFF
--- a/src/main/java/com/o19s/es/ltr/query/RankerQuery.java
+++ b/src/main/java/com/o19s/es/ltr/query/RankerQuery.java
@@ -174,6 +174,11 @@ public class RankerQuery extends Query {
                 public Scorer scorer(LeafReaderContext context) throws IOException {
                     return new ConstantScoreScorer(this, score(), DocIdSetIterator.all(context.reader().maxDoc()));
                 }
+
+                @Override
+                public boolean isCacheable(LeafReaderContext ctx) {
+                    return true;
+                }
             };
         }
         List<Weight> weights = new ArrayList<>(queries.size());
@@ -185,6 +190,13 @@ public class RankerQuery extends Query {
 
     public class RankerWeight extends Weight {
         private final List<Weight> weights;
+
+        @Override
+        public boolean isCacheable(LeafReaderContext ctx) {
+            // Only cachable if model doesn't change?
+            // I *think* it should be the case it should be true
+            return true;
+        }
 
         RankerWeight(List<Weight> weights) {
             super(RankerQuery.this);
@@ -294,10 +306,10 @@ public class RankerQuery extends Query {
                 return ranker.score(fv);
             }
 
-            @Override
-            public int freq() throws IOException {
-                return scorers.size();
-            }
+//            @Override
+//            public int freq() throws IOException {
+//                return scorers.size();
+//            }
 
             @Override
             public DocIdSetIterator iterator() {
--- a/src/main/java/com/o19s/es/ltr/query/RankerQuery.java
+++ b/src/main/java/com/o19s/es/ltr/query/RankerQuery.java
@@ -193,8 +193,10 @@ public class RankerQuery extends Query {
 
         @Override
         public boolean isCacheable(LeafReaderContext ctx) {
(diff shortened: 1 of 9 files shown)


src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -186,6 +191,13 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
     public class RankerWeight extends Weight {
         private final List<Weight> weights;
 
+        @Override
+        public boolean isCacheable(LeafReaderContext ctx) {
+            // Only cachable if model doesn't change?
+            // I *think* it should be the case it should be true
+            return true;
```
nomoa: I'd go with the BooleanWeight approach here: https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/BooleanWeight.java#L314

jettro:
It seems the freq() method of Scorer has been removed from the parent, but some scorers still have it. TermScorer and ExactPhraseScorer for instance. Not sure why yet. But here is the Lucene commit that makes the change:
https://github.com/apache/lucene-solr/commit/183571c085201a2f1e126852647910d1930d50a6

Also not su much a solution, more part of an explanation:
https://issues.apache.org/jira/browse/LUCENE-6278

Can we maybe use the elastic endpoint term vectors here?
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-termvectors.html

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
nomoa: So checked a bit more what's happening here:
- the valid way is to delegate to sub weights like what is done the BooleanWeight

But in fact it's just called to see if lucene can cache the docIds if this query was used as a filter.
So could very much just `throw new IllegalStateException("RankerWeight should not be used as a filter and cached");`

nomoa:
For the freq() problem I'd go with the easy way and check if the child scorer is a TermScorer instance then cast so that you can access the freq() method there as it's still available (as suggested by jettro).

The hard way (we'll have to do it if they remove freq() from TermScorer) would be:
- instead of accepting an arbitrary query we'll accept a query string with field or an analyzer
- tokenize the query based on the analyzer provided
- Scan the freq() from the postings

Using termVector here is not option imo, it'd be too slow.

softwaredoug:
@nomoa I think if we do that (w/ freq), we're officially supporting only the [ES term query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html) and we're assuming Elasticsearch will always map their term query to Lucene's term query?

softwaredoug:
Looks like it may not be possible to check if scorer is a TermScorer, TermScorer appears to be [package private](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/TermScorer.java#L29)

nomoa:
arg... if TermScorer is package private I don't have easy solution to suggest... :(
Going the hard way is much more code to write...

nomoa:
LGTM on my side, the build pass locally.
CI needs to be upgraded to java9.

softwaredoug:
Upgraded Circle to 2.0 to get a Java 9 (a yak shave, within a yak shave :) ). Seems to work

nomoa:
LGTM, I'd just update the PR to remove the WIP flag
The last failure seems to be related to a JVM bug, if possible I'd check if a hs_err_pidXXX.log was generated somewhere but not sure that CircleCI keeps an image of the FS somewhere of the failed builds.

purbon:
test pass for me locally (mac os x)
screen shot 2018-03-20 at 15 20 41

can we retrigger the build in cicle CI and join the build with ssh to try to investigate the error a bit more?

softwaredoug:
Retriggering didn't work, following up on the circle support forums. It appears to be related to a JVM not having enough memory.

softwaredoug:
Build eventually passed...

PR #156, opened 2018-05-15
Selective features for models at query time

umeshdangat:
Some Yelp search flows currently run linear models within a homegrown elasticsearch plugin. The entire model with features is shoved in as `params` to our plugin which `knows` how to extract features and score the individual components. I was investigating trying to use the LTR plugin for this purpose, since it also allows us to use other kinds of models (e.g. xgboost, ranklib) using the same framework and provides options of getting rid of some of our custom ser-de code. However there are some quirks we have before we could port our linear model to being used with LTR.

In our current approach, we train 'N' ~500 features for the model. Then at query time we only use a subset of these features and query elasticsearch using them.  While trying port to using LTR plugin, I realized we need to add some options to it like being able to pass selective features and then score only using them since we do not have all of them available at query time.

My current approach  tries to use the Lucene `MatchNoDocsQuery` query for such nonActive features and the scorer simply returns null for these in `RankerScorer`.  

sltr query would look like below. We only specify the 
> params for active_features and
> active_features

```
"sltr": {
    "model": "my_example_model",
     "params": { "feature_two_brunch_query", "feature_two_brunch_language", "feature_two_brunch_clicks", "feature_two_brunch_impressions"},
     "active_features": {"feature_two_brunch": true}
}

```
If this approach makes sense I can write tests and add support in validation queries as well. Here are more detailed [gists](https://gist.github.com/umeshdangat/2524b9418065b2dbeeaf41cebd72d0ff)


nomoa:
I think this makes total sense.

To fully understand would this work if we added a param like `ignore_on_missing_param` to the feature itself.
We could prevent the feature from being parsed if one the params it depends on is missing. Currently I think it simply throws an exception.
I think this might offer better usability since you don't have to make your client aware of the feature names.

purbon:
I like this feature very much, also it makes complete sense for me also.

nomoa:
Oh I just realized that you already pushed some code, let me know what you think about this `ignore_on_missing_param` solution then I'll take a look at your patch.
(Thanks for digging into the code!)

umeshdangat:
@nomoa , please correct me if I misunderstood your comment about using `ignore_on_missing_param`. As far as I can tell `ignore_on_missing_param` would work well if all features had parameters. However some of our features don't take in any parameters. For those cases `queryParams` in the `Feature` will always be empty and `missingParams` will always be empty. So we cannot tell from `missingParams` if a feature is optional or not at query time. In my current code I maintain `Set<String> activeFeature` to decide if feature is active or not.

```
            if (!isActiveFeature(activeFeatures)) {
                return new MatchNoDocsQuery("This query will not be used for scoring");
            }

```

umeshdangat:
I simplified the code by not passing `activeFeatures` all the way to `RankerScorer` since for the disabled features we already have a NoopScorer scorer and the `RankerScorer.score` for loop takes care of not scoring on them.

I still need to add tests. Also I have TODOs in a couple places which relate to `ValidatingLtrQueryBuilder`.  Will look into that if this patch seems an okay way to proceed.

nomoa:
Thanks this greatly simplifies the whole PR.

But before going further would mind taking a look at https://github.com/o19s/elasticsearch-learning-to-rank/commit/4371c8a0def9c58eae32bbe1545301269ba3f974 it is part of the optimization branch I started but haven't had time to finish (https://github.com/o19s/elasticsearch-learning-to-rank/commits/optimizations)

In short I already had the need to pass more context to the whole query building process so instead of adding new method parameters I added a new class `LtrQueryContext` in place of the `QueryShardContext`. This would allow you to store your `activeFeatures` filters without needing to change all the method signatures but also will help further contributions by having a place to store the various states needed when building the query.
Feel to cherry-pick this specific commit.
Thanks!

umeshdangat:
@nomoa  I cherry-picked commit 4371c8a. This makes sense to me. The code is much more succinct now. I tested locally, still need tests though. Let me know what you think.

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -106,6 +109,7 @@ protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeOptionalString(modelName);
         out.writeOptionalString(featureSetName);
         out.writeMap(params);
+        out.writeMap(activeFeatures);
```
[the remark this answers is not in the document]
ebernhardson: gating this on the elasticsearch version doesn't seem right, as the important bit is the version of the plugin.  It seems plausible the code gets backported to another es version. Maybe YAGNI though i'm not sure.  Unfortunately i also have no suggestions for how to gate this properly on our plugin versions. The only thing i can think of would have required us storing a version header on the actual serialized objects, and that would still be iffy.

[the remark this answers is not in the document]
nomoa: true, sadly I see no other options for now. This kind of version check basically prevents backports `Version.V_6_2_3` will trigger compilation errors. 
We might consider wrapping all the streams with our own Stream[In/Out]put when moving to 7 so that we can store the plugin version as you suggest allowing perhaps to backport more easily.

As for this specific PR we can perhaps do what is wanted without changing the binary format.
If we agree on a *magic name* we could store this feature filter inside a param of the sltr query itself. A name like `__ltr_active_features__`.
The query would look like:
```json
"sltr": {
  "params": {
         "query_string": "user query",
         "random_param": { },
         "__ltr_active_features__": [ "feat1", "feat2" ],
  },
  "model": "foobar"
}
```

I'm perfectly fine with this solution assuming it's well documented.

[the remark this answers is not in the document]
nomoa: fwiw this PR embarks the version bump to 6.2.3 so this is a guarantee that this plugin is not deployed anywhere on es 6.2.3 without this code.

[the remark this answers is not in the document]
nomoa: oops I see 6.2.3 in http://es-learn-to-rank.labs.o19s.com/ this means a release have been made without a commit on the master branch :(
In other words this version check will prevent users using this plugin on 6.2.3 to do a rolling upgrade.
I might suggest to use the magic param solution or bump to another elastic version.

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -68,6 +69,7 @@
     private String featureSetName;
     private String storeName;
     private Map<String, Object> params;
+    private Map<String, Object> activeFeatures;
```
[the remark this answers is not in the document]
umeshdangat: Not really. Fixed.

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -80,6 +82,7 @@ public StoredLtrQueryBuilder(FeatureStoreLoader storeLoader, StreamInput input)
         modelName = input.readOptionalString();
         featureSetName = input.readOptionalString();
         params = input.readMap();
+        activeFeatures = input.readMap();
```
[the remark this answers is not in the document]
umeshdangat: Updated to use elasticsearch version 6.2.4 or after

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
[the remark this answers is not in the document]
umeshdangat: Updated to use elasticsearch version 6.2.4 or after.

umeshdangat:
- Added tests
- Used es 6.2.4 
- Added check to throw Exception when a non-existent feature name is given to activeFeatures
- Changed activeFeatures to be a list instead of a Map

In response to your comment
> I suggest that you squash all the current commits into a single one, it'll be clearer for the next round of review once you have some tests ready.
I did squash all my commits, although that means now it might be difficult to see the above changes.

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -72,28 +72,37 @@ private RankerQuery(List<Query> queries, FeatureSet features, LtrRanker ranker)
...
      * @return the lucene query
      */
-    public static RankerQuery build(LtrModel model, QueryShardContext context, Map<String, Object> params) {
+    public static RankerQuery build(LtrModel model, LtrQueryContext context, Map<String, Object> params) {
         return build(model.ranker(), model.featureSet(), context, params);
     }
 
-    private static RankerQuery build(LtrRanker ranker, FeatureSet features, QueryShardContext context, Map<String, Object> params) {
+    private static RankerQuery build(LtrRanker ranker, FeatureSet features, LtrQueryContext context, Map<String, Object> params) {
+        validateActiveFeatures(features, context);
```
umeshdangat: added validation of active features to be a subset of overall features here

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -72,28 +72,37 @@ private RankerQuery(List<Query> queries, FeatureSet features, LtrRanker ranker)
...
         List<Query> queries = features.toQueries(context, params);
         return new RankerQuery(queries, features, ranker);
     }
 
+    private static void validateActiveFeatures(FeatureSet features, LtrQueryContext context) {
+        for (String featureName : context.getActiveFeatures()) {
+            if (!features.hasFeature(featureName)) {
+                throw new IllegalArgumentException("Feature : [" + featureName + "] " + "does not exist");
```
nomoa: nitpick: it feels a bit late to do this here, I mean it's perhaps preferable to do the check earlier so that you can tell the user that the missing feature is part of the `active_features` param in the sltr query.
Personally I'd move this check in `StoredLtrQueryBuilder#doToQuery` method.

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -106,6 +117,9 @@ protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeOptionalString(modelName);
         out.writeOptionalString(featureSetName);
         out.writeMap(params);
+        if (out.getVersion().onOrAfter(Version.V_6_2_4)) {
+            out.writeOptionalStringArray(activeFeatures != null ? activeFeatures.toArray(new String[0]) : null);
```
nomoa: nitpick: `activeFeatures.toArray(new String[activeFeatures.size()])`

src/main/java/com/o19s/es/ltr/feature/store/PrecompiledTemplateFeature.java
```diff
@@ -80,36 +81,48 @@ public String name() {
     }
 
     @Override
-    public Query doToQuery(QueryShardContext context, FeatureSet set, Map<String, Object> params) {
+    public Query doToQuery(LtrQueryContext context, FeatureSet set, Map<String, Object> params) {
         List<String> missingParams = queryParams.stream()
                 .filter((x) -> params == null || !params.containsKey(x))
                 .collect(Collectors.toList());
-        if (!missingParams.isEmpty()) {
+        if (context.isFeatureActive(name) && !missingParams.isEmpty()) {
```
nomoa: instead of checking this on every Feature implementation I'd do this check on all the FeatureSet implementations in the toQueries method:
```java
        List<Query> queries = new ArrayList<>(features.size());
        for(Feature feature : features) {
            if (context.isFeatureActive(name)) {
               queries.add(feature.doToQuery(context, this, params));
            } else {
               queries.add(new MatchNoDocsQuery( "Feature " + feature.name() + " deactivated" ));
            }
        }
        return queries;
```

The reason is that it's likely we'll add more Feature implementation and it'll be easy to forget to do this check in the future.
Adding new FeatureSet implementations seems unlikely.

nomoa:
I forgot:
 - a few lines to explain this feature in the doc would be awesome.
 - you could add adapt a test in `rest-api-spec/test/fstore/80_search_w_partial_models.yml` to use your new param, that would ensure proper validation of the whole stack.
- Adding a new search request using this new param in `com.o19s.es.ltr.query.StoredLtrQueryIT#testFullUsecase` would be awesome.

You did not adapt `ValidatingLtrQueryBuilder`, I'll leave this up to you to decide if it's worth doing it or not.

Thanks!

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
umeshdangat: regarding initializing the array with zero sized element:  seems like, providing a zero-length array, even creating it and throwing it away, is on average faster than allocating an array of the right size as per this blogpost: https://shipilev.net/blog/2016/arrays-wisdom-ancients/

I am okay using your suggestion if this seems like an micro-optimization and you prefer to keep this codebase more consistent with initializing with `size()`

nomoa: interesting read thanks :)
I probably have bad habits so whenever I see something that could be initialized to the right size I point it. But please keep the code as you like and feel free to ignore this nitpick.

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
umeshdangat: done

src/main/java/com/o19s/es/ltr/feature/store/PrecompiledTemplateFeature.java
umeshdangat: Makes sense. done

umeshdangat:
- addressed the last set of PR comments 
- explained this feature in the doc
- added test to `rest-api-spec/test/fstore/80_search_w_partial_models.yml`
- added a search request with `active_features`to `com.o19s.es.ltr.query.StoredLtrQueryIT#testFullUsecase` to make sure scoring is as expected
-

docs/searching-with-your-model.rst
```diff
@@ -51,6 +51,38 @@ Here we execute a query that limits the result set to documents that match "ramb
 
 Viola!
 
+===========================================
+Scoring on a subset of features with `sltr`
```
nomoa: I'd suffix with
> (added in 1.0.1-es6.2.4)

docs/searching-with-your-model.rst
```diff
@@ -51,6 +51,38 @@ Here we execute a query that limits the result set to documents that match "ramb
...
+===========================================
+
+Sometimes you might want to execute your query on a subset of the features rather than use all the ones specified in the model. In this case the features not specified in :code:`active_features` list will not be scored upon. Their score will default to 0.
```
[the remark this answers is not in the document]
umeshdangat: done.

docs/searching-with-your-model.rst
umeshdangat: done.

src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java
```diff
@@ -145,6 +145,14 @@ protected void doXContent(XContentBuilder builder, Params p) throws IOException
         builder.endObject();
     }
 
+    private static void validateActiveFeatures(FeatureSet features, LtrQueryContext context) {
+        for (String featureName : context.getActiveFeatures()) {
+            if (!features.hasFeature(featureName)) {
+                throw new IllegalArgumentException("Feature : [" + featureName + "] " + "does not exist");
```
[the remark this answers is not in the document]
umeshdangat: Changed to 
`Feature: [foo] provided in active_features does not exist`.

Is this what you suggested?

nomoa:
@softwaredoug it's still unclear to me... I think it really depends on how the model works so in my opinion it's up to the model to decide what to do with missing values. For instance we implemented the decision tree ranker in such a way that it does not distinguish between zeros and missing values. For me this problem is still a bit fuzzy... (Perhaps we are still lacking usecases where a difference between missing features and scores equal to 0 make a real difference.)
For this particular PR I think it's fine to mark them as missing and let the model decide what to do with them.

@umeshdangat thanks for the great work, for me it's good to go. I'll leave the PR open few days just in case someone wants to jump in and comment. If no one opposes I'll merge this next Monday.

purbon:
@softwaredoug @nomoa would that PR deserve new release of the plugin? I kinda think so...

PR #170, opened 2018-06-25
poc add derived feature support in native plugins. reuses poc from sha: d629523

umeshdangat:
@nomoa thanks for the POC!

I added some wiring code on top of your [poc: d629523  ](https://github.com/nomoa/elasticsearch-learning-to-rank/commit/d629523d5d29f60a45c7dd5f7ede5844430c84ea) to get this code working end to end.

Maybe I am missing something but although this seems to work, in its current form the native java plugins still seem to have have a compile time dependency on LTR  and need to access the `FeatureVector` with something like:

`((ScriptFeature.SupplierSupplier)queryParams.get("feature_vector")).get().getFeatureScore(0)`

We can make this simpler and use `featureName` like
`((ScriptFeature.SupplierSupplier)queryParams.getFeatureScore(featureName)`

but the native java code still seems to depend on `SupplierSupplier` class from LTR. 

Is there a way we could avoid and this and the native plugin just use a Map<String, double> instead which is provided by `Supplier`?

### DIFF
--- a/src/main/java/com/o19s/es/ltr/feature/store/FeatureSupplier.java
+++ b/src/main/java/com/o19s/es/ltr/feature/store/FeatureSupplier.java
@@ -1,0 +1,121 @@
+package com.o19s.es.ltr.feature.store;
+
+import com.o19s.es.ltr.feature.FeatureSet;
+import com.o19s.es.ltr.ranker.LtrRanker;
+
+import java.util.AbstractMap;
+import java.util.AbstractSet;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.function.Supplier;
+
+
+public class FeatureSupplier extends AbstractMap<String, Float> implements Supplier<LtrRanker.FeatureVector> {
+    private Supplier<LtrRanker.FeatureVector> vectorSupplier;
+    private final FeatureSet featureSet;
+
+    FeatureSupplier(FeatureSet featureSet) {
+        this.featureSet = featureSet;
+    }
+
+    @Override
+    public LtrRanker.FeatureVector get() {
+        return vectorSupplier.get();
+    }
+
+    public void set(Supplier<LtrRanker.FeatureVector> supplier) {
+        this.vectorSupplier = supplier;
+    }
+
+    /**
+     * Returns {@code true} if this map contains a mapping for the specified
+     * featureName.
+     *
+     * @param featureName featureName whose presence in this map is to be tested
+     * @return {@code true} if this map contains a mapping for the specified
+     * featureName
+     * @throws ClassCastException if the key is of an inappropriate type for
+     *                            this map
+     */
+    @Override
+    public boolean containsKey(Object featureName) {
+        return featureSet.hasFeature((String) featureName);
+    }
+
+    /**
+     * Returns the score to which the specified featureName is mapped,
+     * or {@code null} if this map contains no mapping for the featureName.
+     *
+     * @param featureName the featureName whose associated score is to be returned
+     * @return the score to which the specified featureName is mapped, or
+     * {@code null} if this map contains no mapping for the key
+     */
+    @Override
+    public Float get(Object featureName) {
+        int featureOrdinal;
+        try {
+            featureOrdinal = featureSet.featureOrdinal((String) featureName);
+        } catch (IllegalArgumentException e) {
+            return null;
+        }
+        return vectorSupplier.get().getFeatureScore(featureOrdinal);
+    }
+
+    /**
+     * Strictly speaking the only methods of this {@code Map} needed are
+     * containsKey and get. The remaining methods help fix issues like
+     * - deserialization of FEATURE_VECTOR parameter as a Map object
+     * - keeps editors like intellij happy when debugging e.g. providing introspection into Map object.
+     */
+
+    @Override
+    public Set<Entry<String, Float>> entrySet() {
+        return new AbstractSet<Entry<String, Float>>() {
+            @Override
+            public Iterator<Entry<String, Float>> iterator() {
+                return new Iterator<Entry<String, Float>>() {
+                    private int index;
+
+                    @Override
+                    public boolean hasNext() {
+                        LtrRanker.FeatureVector featureVector = getFeatureVector();
+                        if (featureVector != null) {
+                            return index < featureSet.size();
+                        }
+                        return false;
+                    }
+
+                    @Override
+                    public Entry<String, Float> next() {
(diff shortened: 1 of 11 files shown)


src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
```diff
@@ -0,0 +1,193 @@
...
+
+            @Override
+            public boolean isCacheable(LeafReaderContext ctx) {
+                return false;
+            }
+        }
+    }
+
+
+    static class SupplierSupplier implements Supplier<LtrRanker.FeatureVector> {
```
nomoa: I think we can fix the runtime dependency by implementing `Map<String, Float>` here.
A bunch of methods will return throw new UnsupportedOperationException but `get(String featureName)` and `containsKey(String featureName)` can be implemented and should be all what we need.

umeshdangat: @nomoa implementing `Map` seems to work. If you think this approach makes sense I can work on adding some tests and maybe more cleanup of code.

nomoa: @umeshdangat great!
yes I think the code deserves some cleanup (esp. better naming: SupplierSupplier is really not the most obvious class name :).
For the tests I think it'd be interesting to test : 
- a custom java script plugin (You can inject your own test plugin)
- possibly a painless script feature that uses a couple features exposed through this map

as for usability we have the same problem that we have for derived expressions, users have to declare these kind of features after the dependent features, e.g. if the script is declared at ordinal 1 but the feature it depends on is declared at ordinal 3 it won't work. I have no clue how to make this more obvious for users.

Thanks!

umeshdangat:
@nomoa I changed `FeatureSupplier` to extend an `AbstractMap`. This allows our implementation to be more complete and helps with things like 

>           - deserialization of FEATURE_VECTOR parameter as a Map object using standard java libraries like jackson. Initially the Map is empty.
>          - keeps editors like intellij happy when debugging e.g. providing introspection into Map object.

This meant I had to add a `size()` method on `FeatureVector` interface to implement `hasNext` and `next` on the Map Iterator 

For tests, I added a new java plugin called `NativeScriptPlugin` and this is invoked in existing `StoredLtrQueryIT`  to test end to end functionality.

umeshdangat:
@nomoa could you please take a look at my latest changes?

src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
```diff
@@ -240,63 +244,62 @@ public Float get(Object featureName) {
...
+                            return null;
+                        }
+                    };
+                }
+
+                @Override
+                public int size() {
+                    LtrRanker.FeatureVector featureVector = getFeatureVector();
+                    if (featureVector != null) {
+                        return featureVector.size();
```
nomoa: featureSet.size()

src/main/java/com/o19s/es/ltr/ranker/LtrRanker.java
```diff
@@ -60,5 +60,10 @@
          * Get the feature score
          */
         float getFeatureScore(int featureId);
+
+        /**
+        * Get the size of this FeatureVector
+        */
+        int size();
```
nomoa: perhaps not needed since you have the featureSet available

src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
```diff
@@ -0,0 +1,305 @@
...
+
+            @Override
+            public boolean isCacheable(LeafReaderContext ctx) {
+                return false;
+            }
+        }
+    }
+
+
+    static class FeatureSupplier extends AbstractMap<String, Float> implements Supplier<LtrRanker.FeatureVector> {
```
nomoa: perhaps we should extract this class into its own file.
Having a dedicated unit test for it would be nice I think.

umeshdangat: fixed, still need to add unit tests though.

src/main/java/com/o19s/es/ltr/ranker/LtrRanker.java
umeshdangat: good point, sorry for the oversight. fixed.

umeshdangat:
Thanks @nomoa ! Sorry for the delayed response from my end to. Still need to write some tests, and documents. Will do that next.

umeshdangat:
@nomoa 
- Fixed the TODO for `explain`
- Added unit tests for `FeatureSupplier`
- Added documentation for using this feature. I tried adding integration tests for `painless` but we do not currently load painless in the integration cluster. Error: `script_lang not supported [painless]`. I did locally verify that painless does actually load the feature_vector. Also, we do have integration tests for a native plugin `NativeScriptPlugin` in the tests.

src/main/java/com/o19s/es/ltr/feature/store/FeatureSupplier.java
```diff
@@ -0,0 +1,126 @@
...
+     * @param featureName featureName whose presence in this map is to be tested
+     * @return {@code true} if this map contains a mapping for the specified
+     * featureName
+     * @throws ClassCastException if the key is of an inappropriate type for
+     *                            this map
+     */
+    @Override
+    public boolean containsKey(Object featureName) {
+        try {
+            featureSet.featureOrdinal((String) featureName);
```
nomoa: return featureSet.hasFeature(featureName);

src/main/java/com/o19s/es/ltr/feature/store/FeatureSupplier.java
```diff
@@ -0,0 +1,126 @@
...
+     * @param featureName the featureName whose associated score is to be returned
+     * @return the score to which the specified featureName is mapped, or
+     * {@code null} if this map contains no mapping for the key
+     */
+    @Override
+    public Float get(Object featureName) {
+        int featureOrdinal;
+        try {
+            featureOrdinal = featureSet.featureOrdinal((String) featureName);
+        } catch (IllegalArgumentException e) {
```
nomoa: I'd let IllegalArgumentException propagate, it's likely that it'll endup with an obscure NullPointerException in the end. I think it's preferable to see "Unknown feature XYZ".

nomoa:
LGTM, great addition thanks!

left a couple of minor comments
Feel free to merge once that is resolved or as is if you prefer the current approach.

PR #172, opened 2018-07-23
Add support for parsing lightgbm models exported to json

ebernhardson:
* Parses models exported via python lightgbm.Booster.dump_model
* Example model here generated with advanced_example.py from lightgbm examples
* Supports various missing value handling
* Does not support categorical features
* Model type is `model/lightgbm+json`

### DIFF
--- a/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
+++ b/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
@@ -90,12 +90,24 @@ public class NaiveAdditiveDecisionTree extends DenseLtrRanker implements Account
         private final Node right;
         private final int feature;
         private final float threshold;
+        private final Comparer comparer;
+
+        @FunctionalInterface
+        public interface Comparer {
+            static Comparer GT = (t, v) -> t > v;
+            boolean compare(float threshold, float value);
+        }
 
         public Split(Node left, Node right, int feature, float threshold) {
+            this(left, right, feature, threshold, Comparer.GT);
+        }
+
+        public Split(Node left, Node right, int feature, float threshold, Comparer comparer) {
             this.left = Objects.requireNonNull(left);
             this.right = Objects.requireNonNull(right);
             this.feature = feature;
             this.threshold = threshold;
+            this.comparer = comparer;
         }
 
         @Override
@@ -109,7 +121,7 @@ public class NaiveAdditiveDecisionTree extends DenseLtrRanker implements Account
             while (!n.isLeaf()) {
                 assert n instanceof Split;
                 Split s = (Split) n;
-                if (s.threshold > scores[s.feature]) {
+                if (comparer.compare(s.threshold, scores[s.feature])) {
                     n = s.left;
                 } else {
                     n = s.right;
--- a/src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
+++ b/src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
@@ -1,0 +1,275 @@
+package com.o19s.es.ltr.ranker.parser;
+
+import com.o19s.es.ltr.feature.FeatureSet;
+import com.o19s.es.ltr.ranker.dectree.NaiveAdditiveDecisionTree;
+import com.o19s.es.ltr.ranker.dectree.NaiveAdditiveDecisionTree.Split;
+
+import org.elasticsearch.common.ParseField;
+import org.elasticsearch.common.ParsingException;
+import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
+import org.elasticsearch.common.xcontent.NamedXContentRegistry;
+import org.elasticsearch.common.xcontent.ObjectParser;
+import org.elasticsearch.common.xcontent.XContentParser;
+import org.elasticsearch.common.xcontent.json.JsonXContent;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class LightGBMJsonParser implements LtrRankerParser {
+    public static final String TYPE = "model/lightgbm+json";
+    enum MissingType {
+        None, Zero, NaN;
+    }
+
+    @Override
+    public NaiveAdditiveDecisionTree parse(FeatureSet set, String model) {
+        RootParserState root;
+        List<NaiveAdditiveDecisionTree.Node> trees;
+        try (XContentParser parser = JsonXContent.jsonXContent.createParser(NamedXContentRegistry.EMPTY, LoggingDeprecationHandler.INSTANCE, model)) {
+            trees = RootParserState.parse(parser, set).toNodes(set);
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Cannot parse model", e);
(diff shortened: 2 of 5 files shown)


src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
```diff
@@ -109,7 +121,7 @@ public float eval(float[] scores) {
             while (!n.isLeaf()) {
                 assert n instanceof Split;
                 Split s = (Split) n;
-                if (s.threshold > scores[s.feature]) {
+                if (comparer.compare(s.threshold, scores[s.feature])) {
```
ebernhardson: this wont work actually. I'm not sure if it actually uses it, but the way the lightgbm export is defined the comparer can be different for each split. We should use s.comparer.compare(...), although that might prevent inlining. I should run a benchmark to see if that matters at all.

src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
```diff
@@ -0,0 +1,275 @@
...
+        public static SplitParserState parse(XContentParser parser, FeatureSet set) {
+            SplitParserState split = PARSER.apply(parser, set);
+            if (split.isSplit()) {
+                if (!split.hasAllSplitFields()) {
+                    throw new ParsingException(parser.getTokenLocation(), "This split does not have all the required fields");
+                }
+                if (!"<=".equals(split.decisionType)) {
+                    throw new ParsingException(parser.getTokenLocation(), "Only the <= decision_type is supported");
+                }
+                if (!split.defaultLeft) {
```
ebernhardson: The comparers support defaultLeft == false, this is unnecessary

src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
```diff
@@ -0,0 +1,275 @@
...
+            if (isSplit()) {
+                String featureName = featureNames.get(splitFeature);
+                if (!set.hasFeature(featureName)) {
+                    // Can't throw parsing exceptions anymore...
+                    throw new RuntimeException("TODO");
+                }
+                int ordinal = set.featureOrdinal(featureName);
+                NaiveAdditiveDecisionTree.Node left = leftChild.toNode(featureNames, set);
+                NaiveAdditiveDecisionTree.Node right = rightChild.toNode(featureNames, set);
+                return new Split(left, right, set.featureOrdinal(featureName), threshold, numericalDecision());
```
ebernhardson: already have the ordinal in a variable, no need to ask again (or remove the variable declaration)

src/test/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParserTests.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    private String generateRandomSplit() {
+        String left = random().nextBoolean() ? generateRandomSplit() : generateRandomLeaf();
+        String right = random().nextBoolean() ? generateRandomSplit() : generateRandomLeaf();
+        return generateRandomSplit(left, right);
+    }
+
+    private String generateRandomSplit(String left, String right) {
+        return generateSplit(random().nextFloat(), random().nextInt(), left, right);
```
ebernhardson: nextInt() needs to be constrained to valid feature numbers

src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
```diff
@@ -0,0 +1,275 @@
...
+
+        // This is the isZero check used in lightgbm.
+        static final float ZERO_THRESHOLD = 1e-35F;
+
+        static boolean isZero(float fval) {
+            return fval > -ZERO_THRESHOLD && fval <= ZERO_THRESHOLD;
+        }
+
+        // See lightgbm src/io/tree.cpp Tree::NumericalDecisionIfElse
+        // TODO: NaN's are invalid lucene scores, should we even check?
```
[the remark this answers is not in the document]
nomoa: well actually it depends on what's done on the FeatureVector#reset() method, basically missing values are not set so if the FeatureVector decides to initialize everything to NaN then that it's the missing value indicator...
On the other hand we have no FeatureVector implementations that reset to NaN everything uses 0 int reset() currently.
It's really not obvious to decide where this logic should belong to...

src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
a-johnston: Coming from xgboost and not being very familiar with lightgbm- does lightgbm differentiate between `>=` and `>`? If not, would be possible in the case of `<`/`<=` to just swap left and right? It seems to make sense that all splits would be between buckets / sample values as opposed to on them. Reading below, I'm not sure how that might interact with the zero special case logic.

Related to the benchmark, the underlying GBT inference implementation here seems to be based on pointer-linked trees. At Yelp, we found pretty significant improvements from preordering the tree nodes in an array ([code](https://github.com/Yelp/xgboost-predictor-java/blob/master/src/main/java/biz/k11i/xgboost/tree/PreorderRegTree.java), [results](http://www.sysml.cc/doc/89.pdf)), but that definitely would be a different issue/PR.

src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
```diff
@@ -0,0 +1,275 @@
...
+        static private final Split.Comparer COMPARER_ZERO_DEFAULT_LEFT = (threshold, fval) -> fval <= threshold || isZero(fval) || Float.isNaN(fval);
+        static private final Split.Comparer COMPARER_ZERO_DEFAULT_RIGHT = (threshold, fval) -> fval <= threshold && !isZero(fval) && !Float.isNaN(fval);
+        static private final Split.Comparer COMPARER_DEFAULT_LEFT = (threshold, fval) -> fval <= threshold || Float.isNaN(fval);
+        static private final Split.Comparer COMPARER_DEFAULT_RIGHT = (threshold, fval) -> fval <= threshold && !Float.isNaN(fval);
+        static private final Split.Comparer COMPARER_SIMPLE = (threshold, fval) -> fval <= threshold;
+
+        public Split.Comparer numericalDecision() {
+            if (MissingType.None.equals(missingType) || (MissingType.Zero.equals(missingType) && defaultLeft && ZERO_THRESHOLD < threshold)) {
```
a-johnston: Seems as though this should be checking against `isZero` instead of just `<`.

src/main/java/com/o19s/es/ltr/ranker/parser/LightGBMJsonParser.java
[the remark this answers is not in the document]
ebernhardson: I think since our underlying stack is lucene and lucene defines scores as positive and non-NaN we should expect the same. reset() can be documented to reset to 0 and that NaN is invalid. This seems like the most obvious mental model and can be easily described.

softwaredoug:
Is this PR still a WIP @ebernhardson ?

ebernhardson:
Closing for now, this is something I expected to have more time to evaluate and move forward, but obviously have not.

PR #173, opened 2018-07-28
movie title encode to support multiple language

wangzheng422:
there are movie title other than english, so add encode to support multiple language.


wangzheng422:
sorry, I mess something unrelated into the PR, I will clean it up.

PR #174, opened 2018-08-04
movie title encode to support multiple language

wangzheng422:
there are movie title other than english, so add encode to support multiple language.


demo/indexMlTmdb.py
```diff
@@ -31,7 +31,7 @@ def bulkDocs(movieDict):
                       "_source": movie}
             yield addCmd
             if 'title' in movie:
-                print("%s added to %s" % (movie['title'], index))
+                print("%s added to %s" % (movie['title'].encode('utf-8'), index))
```
ebernhardson: This doesn't feel like the right place to convert (i'm assuming) a bytes object into a str. The movieDict should be adjusted where it is generated to represent the movie title as a string.

PR #178, opened 2018-08-10
fix profile queries when using FeatureVectorWeight features

umeshdangat:
Elasticsearch has the option to profile queries. However currently this seems to break because of `FeatureVectorWeight` where we choose to throw `UnsupportedOperationException` for `Scorer scorer(LeafReaderContext context)` method and choose to implement `scorer(LeafReaderContext context, Supplier<LtrRanker.FeatureVector> vectorSupplier)` instead.

The issue is incase of `profile:true` elasticsearch wraps our `weight` object inside its own `ProfileWeight` and produces a `ProfileScorer`. This ends up calling `scorerSupplier` which falls back to the base `Weight` class and calls the unsupported `scorer(LeafReaderContext context)` method for our wrapped weight. Here is the code in [elasticsearch](https://github.com/elastic/elasticsearch/blob/99f88f15c5febbca2d13b5b5fda27b844153bf1a/server/src/main/java/org/elasticsearch/search/profile/query/ProfileWeight.java#L66)

I try to add our own implementation for `scorerSupplier` in `FeatureVectorWeight` so we can call our supported scorer instead. One issue here is we need access to `vectorSupplier`. I did this by making Query implement MutableSupplierInterface and making sure we set it in `RankerQuery`.

The following [gist](https://gist.github.com/umeshdangat/ae7bb3a7ecf2c10c52fd0e451a1b75ab) has details of broken profile result to a sample query and the fixed result

### DIFF
--- a/src/main/java/com/o19s/es/ltr/query/FeatureVectorWeight.java
+++ b/src/main/java/com/o19s/es/ltr/query/FeatureVectorWeight.java
@@ -17,10 +17,12 @@
 package com.o19s.es.ltr.query;
 
 import com.o19s.es.ltr.ranker.LtrRanker;
+import com.o19s.es.ltr.utils.Suppliers;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.ScorerSupplier;
 import org.apache.lucene.search.Weight;
 
 import java.io.IOException;
@@ -30,8 +32,12 @@ import java.util.function.Supplier;
  * A weight allowing to inject the FeatureVector in use by the RankerQuery
  */
 public abstract class FeatureVectorWeight extends Weight {
-    protected FeatureVectorWeight(Query query) {
+
+    private final Suppliers.MutableSupplierInterface<Supplier<LtrRanker.FeatureVector>> vectorSupplier;
+
+    protected FeatureVectorWeight(Query query, Suppliers.MutableSupplierInterface<Supplier<LtrRanker.FeatureVector>> vectorSupplier) {
         super(query);
+        this.vectorSupplier = vectorSupplier;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


src/main/java/com/o19s/es/ltr/query/FeatureVectorWeight.java
```diff
@@ -49,10 +55,28 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
...
     /**
-     *
-     * @param context surrent segment
+     * @param context        surrent segment
      * @param vectorSupplier supplier of the feature vector (always call get(), it must not be cached)
      * @return The scorer
      */
     public abstract Scorer scorer(LeafReaderContext context, Supplier<LtrRanker.FeatureVector> vectorSupplier) throws IOException;
+
+    public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
```
nomoa: not entirely sure that this is needed, since the vectorSupplier is now available as a class member.

src/main/java/com/o19s/es/ltr/query/RankerQuery.java
```diff
@@ -245,12 +249,18 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
...
             List<Scorer> scorers = new ArrayList<>(weights.size());
             List<DocIdSetIterator> subIterators = new ArrayList<>(weights.size());
             MutableSupplier<LtrRanker.FeatureVector> vectorSupplier = new MutableSupplier<>();
-            for (Weight weight : weights) {
+            for (int idx = 0; idx < weights.size(); idx++) {
                 Scorer scorer;
+                Query query = queries.get(idx);
+                Weight weight = weights.get(idx);
+                if (query instanceof Suppliers.MutableSupplierInterface) {
+                    ((Suppliers.MutableSupplierInterface<Supplier<LtrRanker.FeatureVector>>) query).set(vectorSupplier);
```
nomoa: not a big fan of mutating the query here...

What I've been experimenting with is a new "ltrRewrite" phase, see https://github.com/nomoa/elasticsearch-learning-to-rank/commit/6cd83de564c95c03ad8daf509d2527995c8cf7ee

nomoa:
closing in favor of #197

PR #180, opened 2018-09-01
allow ltr query param names to be different from source es script par…

umeshdangat:
One thing that I overlooked while writing `ScriptFeature` was that it forces the query_parameter name in LTR script to be same as the source elasticsearch script. For example the painless elasticsearch script below uses `params.multiplier` so the ltr script also needs to use the same name i.e. `multiplier`. It cannot use a different name in params like `multiplier_ltr`.  This change allows us to decouple the naming in source scripts (painless or native) and LTR.

```
{
                  "name": "feature2",
                  "params": [
                    "multiplier_ltr"
                  ],
                  "template_language": "script_feature",
                  "template": {
                    "lang": "painless",
                    "source": "params.feature_vector.get('feature1') * (long)params.multiplier",
                    "params": {
                      "multiplier": "multiplier_ltr"
                    }
                  }
}
```

### DIFF
--- a/src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
+++ b/src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
@@ -39,11 +39,17 @@ public class ScriptFeature implements Feature {
     private final String name;
     private final Script script;
     private final Collection<String> queryParams;
+    private final Map<String, String> ltrToScriptParams;
 
     public ScriptFeature(String name, Script script, Collection<String> queryParams) {
         this.name = Objects.requireNonNull(name);
         this.script = Objects.requireNonNull(script);
         this.queryParams = queryParams;
+        Map<String, String> ltrScriptParams = new HashMap<>();
+        for (Map.Entry<String, Object> entry : script.getParams().entrySet()) {
+            ltrScriptParams.put(String.valueOf(entry.getValue()), entry.getKey());
+        }
+        this.ltrToScriptParams = ltrScriptParams;
     }
 
     public static ScriptFeature compile(StoredFeature feature) {
@@ -84,7 +90,7 @@ public class ScriptFeature implements Feature {
         Map<String, Object> queryTimeParams = new HashMap<>();
         for (String x : queryParams) {
             if (params.containsKey(x)) {
-                queryTimeParams.put(x, params.get(x));
+                queryTimeParams.put(ltrToScriptParams.get(x), params.get(x));
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
```diff
@@ -84,7 +90,7 @@ public Query doToQuery(LtrQueryContext context, FeatureSet featureSet, Map<Strin
         Map<String, Object> queryTimeParams = new HashMap<>();
         for (String x : queryParams) {
             if (params.containsKey(x)) {
-                queryTimeParams.put(x, params.get(x));
+                queryTimeParams.put(ltrToScriptParams.get(x), params.get(x));
```
nomoa: this seems to be breaking change as you now force users to declare all their params as part of this new indirection mechanism
This may as well produce a NPE

src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
```diff
@@ -39,11 +39,17 @@
...
     private final Script script;
     private final Collection<String> queryParams;
+    private final Map<String, String> ltrToScriptParams;
 
     public ScriptFeature(String name, Script script, Collection<String> queryParams) {
         this.name = Objects.requireNonNull(name);
         this.script = Objects.requireNonNull(script);
         this.queryParams = queryParams;
+        Map<String, String> ltrScriptParams = new HashMap<>();
+        for (Map.Entry<String, Object> entry : script.getParams().entrySet()) {
```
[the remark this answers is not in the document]
umeshdangat: I did add a dedicated script params for the indirection.

src/main/java/com/o19s/es/ltr/feature/store/ScriptFeature.java
umeshdangat: adding the dedicated `extra_script_params` should fix this issue

umeshdangat:
@nomoa apologies for the delay in response. I have reworked this PR and incorporated your suggestions for 
- adding tests 
- adding docs around the usage
- using a dedicated `extra_script_params` feature instead of overriding all script.params. 

The basic use case of this feature is as follows:
- We often need to reuse the same underlying native/java script with a fixed set of complex parameters say one of them is a Map<Double, Double> for `weights`. However we can represent 2 different  features in LTR and still reuse the underlying native/java script by simply POSTing two features with different `weights` say `weight_foo` and `weight_bar`. This way during query time we can pass two different set of weights and reuse the underlying native script to compute the score for us.

PR #193, opened 2018-11-06
Applying python coding standards and making the scripts configurable

jettro:
Applied the standard python coding guidelines, which different from the java guidelines. Also added more configuration options to make the scripts easier to use against other environments

### DIFF
--- a/demo/judgments.py
+++ b/demo/judgments.py
@@ -1,23 +1,25 @@
 import re
+from utils import JUDGMENTS_FILE
+
 
 class Judgment:
-    def __init__(self, grade, qid, keywords, docId):
+    def __init__(self, grade, qid, keywords, doc_id):
         self.grade = grade
         self.qid = qid
         self.keywords = keywords
-        self.docId = docId
-        self.features = [] # 0th feature is ranklib feature 1
+        self.docId = doc_id
+        self.features = []  # 0th feature is ranklib feature 1
 
     def __str__(self):
         return "grade:%s qid:%s (%s) docid:%s" % (self.grade, self.qid, self.keywords, self.docId)
 
-    def toRanklibFormat(self):
-        featuresAsStrs = ["%s:%s" % (idx+1, feature) for idx, feature in enumerate(self.features)]
+    def to_ranklib_format(self):
+        features_as_strs = ["%s:%s" % (idx+1, feature) for idx, feature in enumerate(self.features)]
         comment = "# %s\t%s" % (self.docId, self.keywords)
-        return "%s\tqid:%s\t%s %s" % (self.grade, self.qid, "\t".join(featuresAsStrs), comment)
+        return "%s\tqid:%s\t%s %s" % (self.grade, self.qid, "\t".join(features_as_strs), comment)
 
 
-def _queriesFromHeader(lines):
+def _queries_from_header(lines):
     """ Parses out mapping between, query id and user keywords
         from header comments, ie:
         # qid:523: First Blood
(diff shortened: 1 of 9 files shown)


jettro:
There is one thing I'd still want to do, that is changing the file name to remove the camel case and use the more python accepted with underscores.

demo/judgments.py
```diff
@@ -50,27 +53,33 @@ def _judgmentsFromBody(lines):
...
 
+    print(len(argv))
 
+    if len(argv) > 1 and len(argv[1]) > 0:
+        judgement_file_name = argv[1]
+    else:
+        judgement_file_name = JUDGMENTS_FILE
+
+    for judgment in judgments_from_file(judgement_file_name):
+        print(judgment)
```
softwaredoug: In the future, I wonder if we should switch some of these prints to a logging library output?

demo/loadFeatures.py
```diff
@@ -1,48 +1,55 @@
...
-def getFeature(ftrId):
-    return json.loads(open('%s.json' % ftrId).read())
+# Each feature is stored in a file with the name of the file as feature_number.json: 1.json, 2.json. This function
+# load the file from the filesystem using the basepath from the configuration file.
+def get_feature(ftr_id: int):
+    return json.loads(open(BASEPATH_FEATURES + '%s.json' % ftr_id).read())
 
-def eachFeature():
+
+# Find all available features using the pattern 1.json, 2.json, etc. All features have to be in order as we stop the
```
softwaredoug: Thanks for the comments - should these be doc strings instead?

demo/settings.cfg
```diff
@@ -2,3 +2,9 @@
 ESHost = http://localhost:9200
 //ESUser = ltr_demo
 //ESPassword = changeme
+RanklibJar = RankLib-2.8.jar
```
softwaredoug: We might want a version of RankyMcRankFace, as it has output of feature importance

demo/train.py
```diff
@@ -1,27 +1,30 @@
...
+from loadFeatures import init_default_store, load_features
+from utils import elastic_connection, ES_HOST, ES_AUTH, JUDGMENTS_FILE, INDEX_NAME, JUDGMENTS_FILE_FEATURES, \
+    FEATURE_SET_NAME
 
-def trainModel(judgmentsWithFeaturesFile, modelOutput, whichModel=6):
+
+def train_model(judgments_with_features_file, model_output, which_model=6):
     # java -jar RankLib-2.6.jar -ranker 6 -train sample_judgments_wfeatures.txt -save model.txt
-    cmd = "java -jar RankLib-2.8.jar -ranker %s -train %s -save %s -frate 1.0" % (whichModel, judgmentsWithFeaturesFile, modelOutput)
+    cmd = "java -jar RankLib-2.8.jar -ranker %s -train %s -save %s -frate 1.0" % \
```
softwaredoug: Should this use the ranklib jar defined in utils.py?

demo/loadFeatures.py
jettro: Good point, I forgot about those, will move the comments to the right location.

demo/judgments.py
jettro: Yes, that is a good one, will have a look. Should not be too hard.

softwaredoug:
FYI @jettro I merged a rather minor change into master around demos, hence the conflict

jettro:
No problem, was a known problem :-), what do you think, need more changes? I think I resolved all your comments. I can, of course, resolve the conflict. Are we ready for merge then?

PR #195, opened 2018-11-17
WIP - Update to Elasticsearch 6.5 

softwaredoug:
Update of LTR plugin for ES 6.5

- ESLoggerFactory replaced by LogManager
- Defined our own scripting context

TODO 

- [ ] right thing to do for mustache scripts for script context, current solution shouldn't be merged in...


src/main/java/com/o19s/es/ltr/feature/store/StoredFeature.java
```diff
@@ -212,8 +212,8 @@ public Query doToQuery(LtrQueryContext context, FeatureSet set, Map<String, Obje
         // XXX: we hope that in most case users will use mustache that is embedded in the plugin
         // compiling the template from the script engine may hit a circuit breaker
         // TODO: verify that this actually works, it does not feel right
-        ExecutableScript script = context.getQueryShardContext().getScriptService().compile(new Script(ScriptType.INLINE,
-                templateLanguage, template, params), new ScriptContext<>("search", ExecutableScript.class));
+        RankLibScriptEngine.RankLibModelContainer script = context.getQueryShardContext().getScriptService().compile(new Script(ScriptType.INLINE,
```
softwaredoug: We probably need a different script context for the mustache templates. We shouldn't piggyback on the RanklibScriptEngine stuff...

nomoa: it now seems impossible to run a "template" engine in a generic way with the new signatures...
On the other hand I'm pretty sure no one has ever entered this code path as I'm not aware of any other template engine available in elastic. So rather than trying to maintain unreachable code I'd be for considering that anything entering this code path is an "script query" and not a template.

```
        ScriptQueryBuilder builder = new ScriptQueryBuilder(new Script(ScriptType.INLINE, templateLanguage, template, params));
        try {
            return builder.toQuery(context.getQueryShardContext());
        } catch (IOException | ParsingException | IllegalArgumentException e) {
            // wrap common exceptions as well so we can attach the feature's name to the stack
            throw new QueryShardException(context.getQueryShardContext(), "Cannot create query while parsing feature [" + name + "]", e);
        }
```

PR #198, opened 2018-11-28
fix a demo script's name and demo instructions

yasufumi0410:
I found some mistake in demo instructions and demo script's name.
This PR fixes those.


jettro:
Good one, it seems I completely forgot to update the readme file.

PR #204, opened 2018-12-20
Custom rescore context

nomoa:
Add a custom rescore component "ltr_rescore". Meant to enable bulk scoring
provided that the ranker implements BulkLtrRanker but also to provide a
more convenient way to "replace" the query score.

Bulk scoring works by collecting feature values on per feature/segment basis
into a FeatureMatrix. Batch size can be controlled by the ranker but also
overridden by the user providing "scoring_batch_size".

This rescore component provides the same functionalaties as the upstream
rescore but with a new score mode: replace.
Replacing query scores require normalizing the scores into coherent ranges
accross shards. Added the following normalizers:

- minmax: normalizes to [0,1] given a min, max range
- saturation: normalizes to [0,1] given k and a (only works for positive values)
- logistic: normalizes to [0,1] given k and x0
- interval: normalizes to a custom range

Strategy here is e.g. to force query scores to be in the [0,1] ranges and
rescored docs in the [1,2] range.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -46,6 +46,7 @@ Contents
    logging-features
    training-models
    searching-with-your-model
+   ltr-rescoring
    x-pack
    advanced-functionality
    :caption: Contents:
--- a/docs/ltr-rescoring.rst
+++ b/docs/ltr-rescoring.rst
@@ -1,0 +1,185 @@
+Rescoring using the LTR rescorer
+********************************
+
+=====
+Usage
+=====
+
+The :code:`ltr_rescore` is a custom rescorer very similar to the :code:`query_rescorer` provided by elasticsearch.
+Its main advantage is to provide a `replace` mode but also enables bulk scoring capabilities. ::
+
+    {
+        "window_size": 100,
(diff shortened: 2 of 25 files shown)


ebernhardson:
I wanted to note that we've talked about this patch, and it is interesting but we need to do some more benchmarking to determine it's usefulness. One primary concern is that it is common to rescore a relatively small number of docs (~100), and this batches by lucene segment. Basically I suspect it would be quite common to find that there is only 1 or 2 documents per segment to rescore, at least with the WMF prod use case.

I could certainly see other use cases where this might be more useful though. An index that is built, optimized and replaced, rather than continuously updated, may be able to see a greater benefit from batch scoring.

I intend to look back at my data-parallel JNI based tree evaluator, wire it up to this and see how it runs with more realistic data, time permitting.

softwaredoug:
It's a while later, thought I'd just check it to see if you guys still think its something you're working on?

nomoa:
No I don't have immediate plans to continue working on this.
I'll reopen if we find time to improve perf with batching.

PR #208, opened 2019-01-11
Upgrade elasticsearch to 6.5.4

moliware:
Similar commit to what it was done [here](https://github.com/o19s/elasticsearch-learning-to-rank/commit/3075d1cd015c2b5e9f4c24c069bca75674687879) for upgrading to 6.5.2

We are starting to use the plugin and planning to install it during the next upgrade of our production clusters.

Thank you for developing this plugin :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,12 +6,12 @@ buildscript {
   }
 
   dependencies {
-    classpath "org.elasticsearch.gradle:build-tools:6.5.2"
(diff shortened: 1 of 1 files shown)


softwaredoug:
Thanks @moliware ! (PS if you're an active user, please join us at the [search relevance community slack](http://o19s.com/slack) to discuss the plugin in #es-learn-to-rank channel)

Issue #207, opened 2019-01-10
Script Features Parameters - slows down Elasticsearch queries significantly


richardknox:
Hi,

We have been building a LTR system for our platform and have come across a significant performance decrease using Script Features Parameters as the inputs to our model.

We have a feature set containing 16 features, 14 of which were being generated using an inline painless script at rescore time. 

Under load testing we noticed significantly higher response times (p99 of 187ms) than using indexed features (p99 of 55ms) for a window size of 40, with this increasing to p99 of 55000ms for a window size of 1000 (with indexed features this stayed around the 60ms for p99).

It would be useful for the documentation to warn against using these inline scripts for feature generation where possible, and at a minimum state that there is a performance trade-off to consider.

softwaredoug:
Thanks @richardknox 

Do feel free to submit a documentation PR.

One strategy for isolating performance issues is to issue the query as a boolean query with 16 should clauses. If you still have a performance issue, it's related to the queries themselves. If not, the plugin has an issue.

Also, feel free to join us at [search relevance community slack](http://o19s.com/slack) to discuss this plugin in the #es-learn-to-rank channel.

softwaredoug:
Hi @richardknox is this worth closing now, it seems the script performance was poor independent of the LtR plugin?

richardknox:
Hi @softwaredoug, it does seem to be the case our scripts are the cause and not the plugin itself. Depending on what you guys think, I still feel maybe even a warning regarding using painless could have merit.
Also, @nomoa mentioned that there may still be something going on with script_features as it is doubling our response times but that may be something requiring more investigation.
Happy to go with your feeling!

PR #209, opened 2019-01-22
Adds a warning regarding script feature performance - Issue #207

richardknox:
Added a short warning message regarding potential performance penalties associated with using script features

### DIFF
--- a/docs/building-features.rst
+++ b/docs/building-features.rst
@@ -82,6 +82,8 @@ These are essentially :ref:`derived-features`, having access to the :code:`featu
 
 The custom script has access to the feature_vector via the java `Map <https://docs.oracle.com/javase/8/docs/api/java/util/Map.html>`_ interface as explained in :ref:`create-feature-set`.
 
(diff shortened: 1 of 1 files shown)


docs/building-features.rst
```diff
@@ -82,6 +82,8 @@ These are essentially :ref:`derived-features`, having access to the :code:`featu
 
 The custom script has access to the feature_vector via the java `Map <https://docs.oracle.com/javase/8/docs/api/java/util/Map.html>`_ interface as explained in :ref:`create-feature-set`.
 
+(WARNING script features can cause the performance of your Elasticsearch cluster to degrade, if possible avoid using these for feature generation if you require your queries to be highly performant)
```
umeshdangat: TBH I am not sure if the `script_feature` itself is slow since we have been using them in production for a while now and do not see any massive perf degradations. Ideally I would like to know which parts of the script_features make it slow (if that is the case) rather than a blanket warning which makes it somewhat ambiguous

PR #227, opened 2019-06-21
Fix innert hit query for ltr log

taoyyu:
Fixed #226 for inner hits.
Modified the tests with nested documents/query with inner hits.
Added a new test. The new test will break without my change.

Issue #226, opened 2019-06-14
ltr_log is broken for query with innert_hits


taoyyu:
ltr_log seems broken for query with inner_hits
I got an error like 
> "No query named logged_featureset found"

I can reproduce the issue in the following way.
Test Environment
elasticsearch version: 7.0.1
ltr plugin version: built from master

1. Create test Index with nested fields
`curl -X PUT "localhost:9200/test" -H 'Content-Type: application/json' -d'
{
	"settings": {
		 "number_of_replicas": 0
	},
	"mappings": {
			"properties": {
				"group_name": {"type": "text"},
				"years": {"type": "double"},
				"user": {
					"type": "nested"
				}
			}
		}
}
'`

2. Index a document
`curl -X POST "localhost:9200/test/_doc/1" -H 'Content-Type: application/json' -d'
{
	"group_name": "A",
	"years": 10,
	"user": [
		{
			"name": "B",
			"age": 20
		},
		{
			"name": "C",
			"age": 25
		}
	]
}'`

3. Create a simple LTR featureset
`curl -XPOST "localhost:9200/_ltr/_featureset/test_featureset" -H 'Content-Type: application/json' -d'
{
   "featureset": {
        "features": [
            {
                "name": "years",
                "params": [],
                "template_language": "script_feature",
                "template": {
                    "lang": "painless",
                    "source": "doc['\''years'\''].value"
                }
            }
        ]
   }
}'`

4. Create a simple model
`curl -XPOST "localhost:9200/_ltr/_featureset/test_featureset/_createmodel" -H 'Content-Type: application/json' -d'
{
	"model": {
		"name": "test_model",
		"model": {
			"type": "model/linear",
			"definition": "{\"years\": 1.0}"
		}
 	}
 }'`

5. Run a query with inner_hits
`curl -XPOST "localhost:9200/test/_search" -H 'Content-Type: application/json' -d'
{
	"query": {
		"bool": {
			"must": [
				{
					"nested": {
						"path": "user",
						"query": {
							"bool": {
								"filter": [
									{
										"term": {
											"user.age": 20
										}
									}
								]
							}
						},
						"inner_hits": {
							"sort": [
								{
									"user.age": {"order": "desc"}
								}
							]
						}
					}
				},
				{
					"sltr": {
						"model": "test_model",
						"params": {},
						"_name": "logged_featureset"
					}
				}
			]
		}
	},
	"ext": {
		"ltr_log": {
			"log_specs": [
[...]

PR #239, opened 2019-08-23
Backport: Fix innert hit query for ltr log (#227)

aprudhomme:
We need the fix for issue [#227](https://github.com/o19s/elasticsearch-learning-to-rank/issues/226) backported to the ES 6.3.1 branch to unblock one of our use cases.

The fix applied cleanly and the tests pass.


umeshdangat:
looks good to me

PR #245, opened 2019-09-22
Fix typo from unkown to unknown

richardknox:
Correcting typo, then it is easier to notice immediately in logging etc. when your model has not been deployed correctly.


richardknox:
@softwaredoug please have a look if you can 😄

softwaredoug:
Thanks! Merged, probably won't be in a release until we upgrade to 7.4

PR #247, opened 2019-10-14
Add XGBoost objective parameter

aprudhomme:
There are a number of parameters that can be specified for an XGBoost model. https://xgboost.readthedocs.io/en/latest/parameter.html#learning-task-parameters

The 'objective' learning parameter can have an effect on the prediction from a model. For example, a model with the logistic regression objective would return the sigmoid of the model score.

Though this transform would not affect the final ranking order, providing this consistency is helpful when migrating existing models to the ltr plugin. Also (and perhaps more importantly), this provide the ability to avoid producing negative scores.

This branch allows for the specification of a parameter block prepended to the XGBoost model definition. The block is optional and maintain the status quo behavior by default. Only the 'objective' option is currently available, but other parameters can be added as needed.


aprudhomme:
Not sure why the build failed. Passes fine for me https://travis-ci.com/aprudhomme/elasticsearch-learning-to-rank/builds/132034153

softwaredoug:
Thanks @aprudhomme - weird, let me pull it down and try your branch locally.

docs/training-models.rst
```diff
@@ -105,9 +115,26 @@ Or an xgboost model::
...
+
+Or an xgboost model with parameters::
+
+    POST _ltr/_featureset/more_movie_features/_createmodel
+    {
+        "model": {
+            "name": "my_xgboost_model",
+            "model": {
+                "type": "model/xgboost+json",
+                "definition": "{ \"objective\": \"reg:logistic\" }
```
nomoa: this does not seem to be a valid JSON.
Since XGBoost does not seem to define a clear json schema for its model and the only doc I could find is the list of split definitions directly I suggest to start defining one.

The XGBoost parser would have to be able to parse:
- existing format that starts directly with an array of split defitions
- custom format that allows passing params:
```
{
   "objective": "...",
   "splits": [
   ]
}
```

src/main/java/com/o19s/es/ltr/ranker/dectree/ModelObjective.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Map;
+
+/**
+ * Represents an XGBoost model learning objective
+ *
+ * Depending on the objective, the model prediction may require a transform. Currently only untransformed
+ * and logistic (sigmoid) types are implemented.
+ * See <a href="https://xgboost.readthedocs.io/en/latest/parameter.html#learning-task-parameters">task params</a>
+ */
+public class ModelObjective {
```
nomoa: While for XGBoost this is a model objective for the plugin and prediction purpose this is just a normalization of the output score.
I think it would clearer to make this a factory of normalization methods and possibly calling this Normalizers instead of ModelObjective

src/main/java/com/o19s/es/ltr/ranker/dectree/ModelObjective.java
```diff
@@ -0,0 +1,59 @@
...
+    public static final String DEFAULT_OBJECTIVE_TYPE = "reg:linear";
+    private static final Map<String, ModelObjective> OBJECTIVES = new HashMap<>();
+
+    static {
+        register("binary:logistic", new LogisticObjective());
```
nomoa: I would put this map into the XGboost parser as this is really xgboost specific.

aprudhomme:
@nomoa Thank you for the review. I have attempted to address your feedback. Please let me know if this looks better.

src/main/java/com/o19s/es/ltr/ranker/normalizer/Normalizers.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Class that manages Normalizer implementations
+ */
+public class Normalizers {
+    private static final Map<String, Normalizer> NORMALIZERS = new HashMap<>();
+    public static final String DEFAULT_NORMALIZER_NAME = "noop";
+
+    static {
+        register("noop", new NoopNormalizer());
```
nomoa: nit: I think it'd be more "future proof" if we made NORMALIZERS a Collections.immutableMap(), basically only allow static initialization here. I worry of having a mutable Map here.
If we wanted to make Normalizers easier to register I would go the path of a Builder in the LtrPluginConstructor similar to what we do with LtrRankerParserFactory but this seems overkill for this particular patch.

src/main/java/com/o19s/es/ltr/ranker/normalizer/Normalizers.java
```diff
@@ -0,0 +1,55 @@
...
+        register("sigmoid", new SigmoidNormalizer());
```
nomoa: nit: perhaps extract these strings as string constants?

src/main/java/com/o19s/es/ltr/ranker/normalizer/Normalizers.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new IllegalArgumentException(name + " is not a valid Normalizer");
+        }
+        return normalizer;
+    }
+
+    public static boolean exists(String name) {
+        return NORMALIZERS.containsKey(name);
+    }
+
+    public static Normalizer defaultNormalizer() {
```
nomoa: nit: perhaps name it noopNormalizer() ?
having a defaultNormalizer() method here feels like it can be changed while in fact it does not seem to be the responsibility of this class to change the "default" normalizer of the rankers using a Normalizer.

src/main/java/com/o19s/es/ltr/ranker/parser/XGBoostJsonParser.java
```diff
@@ -41,60 +44,109 @@
...
+                definition = new XGBoostDefinition();
+                definition.splitParserStates = new ArrayList<>();
+                while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
+                    definition.splitParserStates.add(SplitParserState.parse(parser, set));
+                }
+            } else {
+                throw new ParsingException(parser.getTokenLocation(), "Expected [START_ARRAY] or [START_OBJECT] but got ["
+                        + startToken + "]");
+            }
+            if (!definition.hasAllFields()) {
```
nomoa: nit: perhaps move this check in the `if (startToken == XContentParser.Token.START_OBJECT) {` and inline the `definition.splitParserStates != null` check so that you can inform in he exception message that the field "splits" that is actually missing.

src/test/java/com/o19s/es/ltr/ranker/parser/XGBoostJsonParserTests.java
```diff
@@ -196,32 +275,6 @@ public void testComplexModel() throws Exception {
...
-        FeatureVector v = tree.newFeatureVector(null);
-        v.setFeatureScore(0, 0.124F);
-        assertEquals(0.2F, tree.score(v), Math.ulp(0.2F));
-        v.setFeatureScore(0, 0.122F);
-        assertEquals(0.5F, tree.score(v), Math.ulp(0.5F));
-        v.setFeatureScore(0, 0.123F);
-        assertEquals(0.2F, tree.score(v), Math.ulp(0.2F));
-    }
-
     private float sigmoid(float f) {
```
nomoa: needed?

src/main/java/com/o19s/es/ltr/ranker/normalizer/Normalizers.java
aprudhomme: Sure, easy registration does seem unnecessary at this point.

src/main/java/com/o19s/es/ltr/ranker/normalizer/Normalizers.java
aprudhomme: I'm thinking maybe just get rid of the default method. The noop normalizer kind of speaks for itself.

src/main/java/com/o19s/es/ltr/ranker/parser/XGBoostJsonParser.java
aprudhomme: I had based this off the split parsing validation. But I guess we can make it more specific for now and reassess if the parsing gets more complicated.

src/test/java/com/o19s/es/ltr/ranker/parser/XGBoostJsonParserTests.java
aprudhomme: It is used in one test, but I can just precompute the values.

Issue #135, opened 2018-02-16
Clarify how we handle missing values, NaN, zeros...


nomoa:
While trying to add support for XGBoost missing direction I realized that the way we handle missing values is not very clear (code&code wise).

During logging we allow users to set `missing_as_zero` which will emit zeros instead of nothing. After that it's up to the user to properly configure its training algorithm to handle these. E.g. XGBoost has support for them and will emit a model with an additional decision `is missing?` besides the threshold check.
Today the model parser for XGBoost completely ignores the missing branch. This basically assumes the features were logged with `missing_as_zero`.
Concerning ranklib, its DataPoint natively supports missing values by using [NaN](https://github.com/o19s/RankyMcRankFace/blob/master/src/ciir/umass/edu/learning/DataPoint.java#L42). But on the plugin glue code we force all values to [zeros](https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/ltr/ranker/ranklib/DenseProgramaticDataPoint.java#L72) on the reset() method (this is actually a subtle bug as the reset method is only called when reusing the feature vector, basically on the first doc we will send NaNs for missing values but zeros on next docs).

Things we should fix regardless:
- when logging we should fail if we emit NaN for a non-missing value (bogus feature/query).
- we should fix [DenseProgramaticDataPoint](https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/ltr/ranker/ranklib/DenseProgramaticDataPoint.java) so that the reset method is coherent with the [class it extends](https://github.com/o19s/RankyMcRankFace/blob/master/src/ciir/umass/edu/learning/DataPoint.java#L42)

Evaluate:
- start to add a `boolean missing(int featureIdx)` for the [FeatureVector](https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/ltr/ranker/LtrRanker.java#L53) interface.
- eventually fix (or add a new impl) our decision implementation so that it supports missing values. Sadly the xgboost format has no way to tell us if the missing branch needs to be checked (optimization).
- doc: clarify how we handle missing values in the various ranker implementations we support so that users can decide properly if they want to log features with `missing_as_zero`.

nomoa:
edit: removed the comment on ranklib DataPoint, the default constructor does nothing so the float array is properly initialized with zeros which is coherent with the reset() method.

[...]

PR #248, opened 2019-10-18
Sparse feature support for xgboost models

akshaykumar90:
We still use a dense feature vector for model evaluation, but reserve `Float.MAX_VALUE` to represent "missing". We then use this value to take the proper node path during model evaluation.

`missingNodeId` was already present in the xgboost json parser but was being ignored. Modified the parser to actually make use of it and send it down to NaiveAdditiveDecisionTree.Split constructor.

The missing node Id has to be one of right or left nodes.

Modified the `simple_tree` format to allow specifying what node to take on missing features. Specified as a boolean—take left on true.

### DIFF
--- a/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
+++ b/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
@@ -88,12 +88,14 @@ public class NaiveAdditiveDecisionTree extends DenseLtrRanker implements Account
         private static final long BASE_RAM_USED = RamUsageEstimator.shallowSizeOfInstance(Split.class);
         private final Node left;
         private final Node right;
+        private final Node onMissing;
         private final int feature;
         private final float threshold;
 
-        public Split(Node left, Node right, int feature, float threshold) {
+        public Split(Node left, Node right, Node onMissing, int feature, float threshold) {
             this.left = Objects.requireNonNull(left);
             this.right = Objects.requireNonNull(right);
+            this.onMissing = onMissing == null ? this.left : onMissing;
             this.feature = feature;
             this.threshold = threshold;
         }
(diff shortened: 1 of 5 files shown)


nomoa:
Adding a new condition to check the split might have a perf impact.
Here you decided to use MAX_VALUE but this is not how missing values are encoded in the rest of the plugin, I suspect this is because you have custom query/feature that follow this convention?
We did not find a consensus yet on how to treat/encode missing values in the plugin please see https://github.com/o19s/elasticsearch-learning-to-rank/issues/135.

I think I would prefer a separate Ranker implementation instead of making the assumption that MAX_VALUE will work everywhere.

akshaykumar90:
@nomoa Thanks for the feedback and pointing out the existing open issue. I would look into the other approach you suggest.

aprudhomme:
@nomoa Yes, we are wrapping some existing features/models and require sparse features evaluation. This was the convention we chose.
It would be best if there was general sparse feature support, but we were trying to do this as a nearer term fix.

Would it be more acceptable to add the 'missing' value as an XGBoost model parameter, like in https://github.com/o19s/elasticsearch-learning-to-rank/pull/247 ?
Providing no value could be interpreted as opting out of missing support, and could be used to avoid the extra condition check during evaluation.

nomoa:
> Would it be more acceptable to add the 'missing' value as an XGBoost model parameter, like in #247 ?
> Providing no value could be interpreted as opting out of missing support, and could be used to avoid the extra condition check during evaluation.

Yes for me it's totally acceptable to have a param in the model definition that instructs some specific behavior of the ranker. Having a way to tell the XGBoost parser what is the missing value it has to check makes sense to me.
In the actual prediction code (NaiveAdditiveDecisionTree) I would try to use this new param as early as possible so that the `is this a missing value?` checks are avoided when not needed.

akshaykumar90:
Closing this pull request. Will open a new one with the proposed changes.

PR #251, opened 2019-10-28
Backport: Add XGBoost objective parameter #247 to ES 6.3.1

aprudhomme:
Backport of https://github.com/o19s/elasticsearch-learning-to-rank/pull/247 to the es_6_3_1 branch.


softwaredoug:
Thanks- generally we wouldn't backport a new feature like this. You can imagine if we cherry picked feature F to version V all the time, then the version and what features are in/out could get difficult to track

Would it be possible for you to do your own build of the 6.3.1 plugin as opposed to having an 'official' build? Let me know if that's doable.

aprudhomme:
Thank you for the reply. After some discussion, we will just use our own unofficial branch with backports. Our applications should hopefully be closer to master soon anyways.

PR #265, opened 2019-12-17
Backport: Fix inner hit query for ltr log

DAVIDFlavien:
We need the fix that has been merged in master for issue #227 backported to the ES 6.8 branch to unblock one of our use cases.

Tests are passing ✅.


softwaredoug:
I likely will be inclined to take the bug-fix backport, but just FYI we're relatively conservative on backports, as it gets hard to remember what PRs go in what ES plugin versions!

For getting unblocked on your own test, are you able to install your locally built version of the plugin? After building locally, you can install the .zip file under `build/distributions/` ie `build/distributions/ltr-1.1.2-es7.4.0.zip`

PR #268, opened 2020-01-07
Add optional extra logging for native scripts

aprudhomme:
We have had some inconvenience with using logging and our native scripts. For some of our feature extractors, we care about not only the final feature value, but also some 'base' feature(s) used/generated during the computation.

Currently, it would be possible to create an LTR feature for each of these 'base' features, then derive the true feature needed for model evaluation. However, this does not work for all of our uses because:
1) We cannot efficiently cache doc value lookups/intermediate computations across features
2) The 'base' features are not necessarily numeric
3) The Scorer interface requires that all these 'base' features be non-negative, which can be challenging

This branch is an attempt to mitigate some of these issues by allowing native scripts to add arbitrary logging information into an extra_logging map. The map is made available during the LoggingFetchSubPhase through a Supplier provided as a parameter.

The map is String -> Object, allowing for structured and non-numeric data. The map is created on first use by a script and results in an additional entry in the hits' logging:
```
{
  "name": "extra_logging",
  "value": <json serialized map>
}
```
If unused, this entry will not appear at all.

Let me know if this approach seems reasonable, or if there is a better way to go about this.

### DIFF
--- a/src/main/java/com/o19s/es/ltr/feature/store/ExtraLoggingSupplier.java
+++ b/src/main/java/com/o19s/es/ltr/feature/store/ExtraLoggingSupplier.java
@@ -1,0 +1,28 @@
+package com.o19s.es.ltr.feature.store;
+
+import com.o19s.es.ltr.ranker.LogLtrRanker;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+
+public class ExtraLoggingSupplier implements Supplier<Map<String,Object>> {
+    protected LogLtrRanker.LogConsumer consumer;
+
+    public void setConsumer(LogLtrRanker.LogConsumer consumer) {
+        this.consumer = consumer;
+    }
+
+    /**
+     * Return a Map to add additional information to be returned when logging feature values.
+     *
+     * This Map will only be non-null during the LoggingFetchSubPhase.
+     */
+    @Override
+    public Map<String, Object> get() {
+        if (consumer != null) {
+            return consumer.getExtraLoggingMap();
+        }
+        return null;
+    }
+}
--- a/src/main/java/com/o19s/es/ltr/feature/store/ExtraLoggingSupplier.java
+++ b/src/main/java/com/o19s/es/ltr/feature/store/ExtraLoggingSupplier.java
@@ -1,16 +1,14 @@
 package com.o19s.es.ltr.feature.store;
 
-import com.o19s.es.ltr.ranker.LogLtrRanker;
-
 import java.util.Map;
 import java.util.function.Supplier;
 
 
 public class ExtraLoggingSupplier implements Supplier<Map<String,Object>> {
-    protected LogLtrRanker.LogConsumer consumer;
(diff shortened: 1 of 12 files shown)


softwaredoug:
Hi @aprudhomme just a note to also add some docs for this in the "advanced" section. And note on the conflicts after I merged in the 7.5 update...

aprudhomme:
Thanks @softwaredoug, I rebased and added a docs section.

softwaredoug:
Hey @nomoa this feature seems like a great idea for advanced users - do you think you could do a review of the code when you get a chance?

nomoa:
Sure! I'll take a look this week.

src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
```diff
@@ -221,6 +224,23 @@ public void accept(int featureOrdinal, float score) {
...
+        /**
+         * Return Map to store additional logging information returned with the feature values.
+         *
+         * The Map is created on first access.
+         */
+        @Override
+        public Map<String,Object> getExtraLoggingMap() {
+            if (extraLogging == null) {
+                extraLogging = new HashMap<>();
+                Map<String,Object> logEntry = new HashMap<>(2);
```
nomoa: nit: not sure it's worthwhile to start with 2 we might rehash since we add two 2 elements? I'd rather use the default initial capacity

src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
```diff
@@ -201,7 +203,7 @@ void doLog(Query query, List<HitLogConsumer> loggers, IndexSearcher searcher, Se
         }
 
         private void rebuild() {
-            List<Map<String, Object>> ini = new ArrayList<>(set.size());
+            List<Map<String, Object>> ini = new ArrayList<>(set.size() + 1);
```
nomoa: nit: add a small comment to explain why you over allocate here, might help the reader to understand that extraLogging will be added when needed

src/main/java/com/o19s/es/ltr/feature/store/ExtraLoggingSupplier.java
```diff
@@ -0,0 +1,28 @@
+package com.o19s.es.ltr.feature.store;
+
+import com.o19s.es.ltr.ranker.LogLtrRanker;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+
+public class ExtraLoggingSupplier implements Supplier<Map<String,Object>> {
```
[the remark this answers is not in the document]
aprudhomme: I'm not sure I will be able to use the MutableSupplier. Since the Map is not created until first access, I will need to store an internal Supplier object rather than a Map object.

aprudhomme:
Thanks @nomoa. I have attempted to address your feedback.

nomoa:
Thanks!

PR #271, opened 2020-01-13
Added model fixr script

jackpf:
Since Lucene 8 does not allow negative scores, we faced a problem that our XGBoost models will cause the LTR plugin to return negative scores.

Our approach was quite simple:
We calculate the most negative leaf value, and append one more tree to the model
which adds the abs() of this smallest value, meaning we always get a positive score,
but relatively the scores will not change.

So here is a PR just in case someone else will find this useful! I wasn't quite sure where this script should go or how it should be instrumented with the LTR plugin, so I'd be happy for any feedback/change requests.


softwaredoug:
Thanks @jackpf! 

Just one note, can we put the name 'xgboost' in the script to clarify it works on xgboost models? Otherwise 👍

jackpf:
Hey @softwaredoug done! :)

softwaredoug:
Awesome, thanks! Finally, just want to confirm, this script is run with Python 3 (like our other demo scripts, etc)?

jackpf:
I tested it with python3 and it worked, I only added `#!/usr/bin/env python` to make it more generic, however I could change it to python3 to be in line with other scripts

softwaredoug:
awesome! Merged! thanks for the contribution!

PR #278, opened 2020-02-13
Closes #269 - Enabling shard request cache

### LINKED ISSUE #269: Request cache not being used
Hi,

It seems like when using the plugin the [shard request cache](https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-request-cache.html#shard-request-cache) is never used.

This is causing massive performance issues in our applications, we generally have a 20% hit ratio in the request cache, which now needs to be computed for every request.

As for my understanding the guilty for not using the cache is [this line](https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/ltr/query/StoredLtrQueryBuilder.java#L158)

When `context.getClient()` method is called , the request cache will be set to disabled.

Would it be possible to improve this behavior in the plugin?

Thanks

iagodvsantos:
# What is it?

As reported by this issue: #269.

Today the plugin is not using the [shard request cache](https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-request-cache.html#shard-request-cache) due to this method call `context.getClient()` on `StoredLtrQueryBuilder` class which makes the Elasticsearch set the request as non-cacheable.

This PR proposes to avoid calling this method on every request so ElasticSearch can cache requisitions.

After simulating some load-tests, we have the following results:

- Plugin not using shard request cache:
Screen Shot 2020-02-13 at 11 53 14

- Plugin using shard request cache:
Screen Shot 2020-02-13 at 11 57 11

Those tests ran against different types of queries, rescoring 5000 documents using an XGboost model. Analyzing the results we can see an improvement of 40% on throughput.

### DIFF
--- a/src/main/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStore.java
+++ b/src/main/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStore.java
@@ -94,12 +94,12 @@ public class IndexFeatureStore implements FeatureStore {
     }
 
     private final String index;
-    private final Client client;
+    private final Supplier<Client> clientSupplier;
     private final LtrRankerParserFactory parserFactory;
 
-    public IndexFeatureStore(String index, Client client, LtrRankerParserFactory factory) {
+    public IndexFeatureStore(String index, Supplier<Client> clientSupplier, LtrRankerParserFactory factory) {
         this.index = Objects.requireNonNull(index);
-        this.client = Objects.requireNonNull(client);
+        this.clientSupplier = Objects.requireNonNull(clientSupplier);
         this.parserFactory = Objects.requireNonNull(factory);
     }
 
@@ -189,7 +189,7 @@ public class IndexFeatureStore implements FeatureStore {
     }
 
     private Supplier<GetResponse> internalGet(String id) {
-        return () -> client.prepareGet(index, ES_TYPE, id).get();
+        return () -> clientSupplier.get().prepareGet(index, ES_TYPE, id).get();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


softwaredoug:
Wow thanks a ton for fixing a long-standing issue @iagodvsantos !

@nomoa I would like this make it into the 7.6 release, do you think you could review this in the next week or so?

src/main/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStore.java
```diff
@@ -94,12 +94,12 @@
     }
 
     private final String index;
-    private final Client client;
+    private final QueryShardContext context;
```
nomoa: we could type this as `Supplier<Client> clientSupplier`, depending on QueryShardContext is not really necessary

src/test/java/com/o19s/es/ltr/action/BaseIntegrationTest.java
```diff
@@ -98,7 +102,12 @@ public FeatureStoreResponse addElement(StorableElement element) throws Execution
     }
 
     public <E extends StorableElement> E getElement(Class<E> clazz, String type, String name, String store) throws IOException {
-        return new IndexFeatureStore(store, client(), parserFactory()).getAndParse(name, clazz, type);
+        return new IndexFeatureStore(store,
```
nomoa: having a Supplier<Client> here would allow to simply switch to:
```return new IndexFeatureStore(store, this::client, parserFactory()).getAndParse(name, clazz, type);```
instead of pulling a fully-featured QueryShardContext

src/main/java/com/o19s/es/ltr/feature/store/index/IndexFeatureStore.java
iagodvsantos: Great! Done!

src/test/java/com/o19s/es/ltr/action/BaseIntegrationTest.java
iagodvsantos: Done!

iagodvsantos:
Could you review it again @nomoa?

nomoa:
thanks!

Issue #155, opened 2018-05-09
Add more metrics to match_explorer


nomoa:
The match_explorer query allows to extract raw index numbers. Currently it allows to extract:
- the number of unique tokens in the query
- stats the term freq in the doc
- stats on document frequency

It could be extended to allow extract more numbers stored in lucene (if available position information for instance).

Please add your suggestions in the comments I'll try to summarize those and update the task description.

li910802:
When we use the match_explorer query, we hope to extract more similarity features between query and documents for training ltr model, such as the startoffset of the first matched term, the proportion of matched terms and total terms. 

For different search circumstances, we need to use different search patterns (match, match_phrase, match_phrase_prefix for instance). So in my opinion it is necessary to extract the accurate feature numbers correspond to different search patterns . But currently passing a match or a phrase query will return the same results.
I wonder if my personal opinion is apporiate.
Thanks!

GerardBouchar:
I think that the proportion of matched terms would be a very interesting feature.

ebernhardson:
The proportion of matched terms is already available, albeit in an indirect way. Basically you can create a feature that counts the number of tokens in the query (using query explorer existing feature set), and you can create another feature that counts the number of matches (bm25 with the right k1 and b parameters will do this).  The existing derived feature functionality can then give you a new feature that is the proportion between those two. I'm not sure if it's better off documenting as a way to use more advanced functionality, or hard coding some particular items that are likely to be useful like % of matched terms.

rtancman:
Hi @nomoa and @softwaredoug!

I saw in project code in [ExplorerQuery.java#L204-L216](https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/explore/ExplorerQuery.java#L204-L216) one way to use `match_explorer` with `sum_raw_tf` creating features this way:

```
POST _ltr/_featureset/some_cool_features
{
  "featureset": {
    "name": "some_cool_features",
    "features": [
      {
        "name": "sum_raw_tf_text",
        "params": [
          "keywords"
        ],
        "template_language": "mustache",
        "template": {
          "match_explorer": {
            "type": "sum_raw_tf",
            "query": {
              "match": {
                "text": "{{keywords}}"
              }
            }
          }
        }
      }
    ]
  }
}
```

[...]

PR #282, opened 2020-03-24
Fix raw_tf in match_explorer

rtancman:
Refer: #155 

When you create some feature using the `raw_tf` sometimes when term does not exist in a document we have a null pointer exception. Example:

```
# create feature
POST _ltr/_featureset/some_cool_features
{
  "featureset": {
    "name": "some_cool_features",
    "features": [
      {
        "name": "sum_raw_tf_text",
        "params": [
          "keywords"
        ],
        "template": {
          "match_explorer": {
            "type": "sum_raw_tf",
            "query": {
              "match": {
                "text": "{{keywords}}"
              }
            }
          }
        }
      }
    ]
  }
}

# test features
curl -H 'content-type:application/json' -X POST http://localhost:9200/*_ltr/_search -d '{
  "_source": ["_ltrlog.main"],
  "query": {
    "bool": {
      "filter": [
        {
          "terms": {
            "_id": [
              "165241254"
            ]
          }
        }
      ],
      "should": [
        {
          "sltr": {
            "_name": "some_cool_features",
            "featureset": "some_cool_features",
            "params": {
              "keywords": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
            }
          }
        }
      ]
    }
  },
  "ext": {
    "ltr_log": {
      "log_specs": {
        "name": "main",
        "named_query": "some_cool_features",
        "missing_as_zero": true
      }
    }
  }
}' | jq
```

### DIFF
--- a/src/main/java/com/o19s/es/explore/ExplorerQuery.java
+++ b/src/main/java/com/o19s/es/explore/ExplorerQuery.java
@@ -247,16 +247,16 @@ public class ExplorerQuery extends Query {
 
             // This impl always scores docs in order, so we can
             // ignore scoreDocsInOrder:
-            return new LalaDefaultBulkScorer(scorer);
+            return new ExplorerDefaultBulkScorer(scorer);
         }
 
-        protected static class LalaDefaultBulkScorer extends BulkScorer {
+        protected static class ExplorerDefaultBulkScorer extends BulkScorer {
             private final Scorer scorer;
             private final DocIdSetIterator iterator;
             private final TwoPhaseIterator twoPhase;
 
             /** Sole constructor. */
-            public LalaDefaultBulkScorer(Scorer scorer) {
+            public ExplorerDefaultBulkScorer(Scorer scorer) {
                 if (scorer == null) {
                     throw new NullPointerException();
                 }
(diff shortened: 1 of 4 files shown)


nathancday:
This looks superb! Thanks for donating your skills to fixing this 🙇

softwaredoug:
@worleydl and @nomoa would you mind having a look - thanks!

src/main/java/com/o19s/es/explore/ExplorerQuery.java
```diff
@@ -232,6 +236,113 @@ public void extractTerms(Set<Term> terms) {
             weight.extractTerms(terms);
         }
 
+        @Override
+        public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
```
nomoa: I don't think we need to implement a custom bulk scorer here.

src/main/java/com/o19s/es/explore/ExplorerScorer.java
```diff
@@ -81,7 +81,11 @@ public int docID() {
 
     @Override
     public DocIdSetIterator iterator() {
-        return subScorer.iterator();
+        try {
+            return subScorer.iterator();
+        } catch (Exception e) {
```
nomoa: we probably don't need this
I suppose you wanted to catch a NullPointerException but I suspect that you already fixed this NPE by returning null directly from ExplorerQuery.ExplorerWeight::scorer instead of creating this scorer

If you added this because of something else we should fix the root cause instead.

nomoa:
@rtancman thanks for the bugfix!

softwaredoug:
Released as 1.2.1 for 7.6.1. Thanks!

PR #283, opened 2020-03-25
Compatibility with ES v7.6.1

thePanz:
Makes the plugin compatible with ES 7.6.1
- [x] Update library dependencies to elasticsearch 7.6.1
- [x] Fix failing test due to missing `RankLibModelContainer::execute()` method


softwaredoug:
Thanks a ton for the upgrade!

Issue #288, opened 2020-03-31
Add term position feature in match_explorer


rtancman:
We saw in [PostingsExplorerQuery.java#L86-L88](https://github.com/o19s/elasticsearch-learning-to-rank/blob/master/src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java#L86-L88) one way to implement term position feature.

Is this important? Because we started to do that ;)

This feature is very important for us and we want to contribute to the project.

worleydl:
Thanks for the PR!

PR #289, opened 2020-04-01
Add term position min, max and avg by documents

rtancman:
Refer: #288

### DIFF
--- a/docs/feature-engineering.rst
+++ b/docs/feature-engineering.rst
@@ -37,6 +37,18 @@ The statistics available include:
 
 Putting the operation and the statistic together, you can see some examples. To get stddev of classic_idf, you would write :code:`stddev_classic_idf`. To get the minimum total term frequency, you'd write :code:`min_raw_ttf`.
 
+Term position statistics
+
+The :code:`type` parameter can be prepended with the operation to be performed across term position for the statistic :code:`max`, :code:`min`, :code:`avg`, and :code:`norma_avg`. The :code:`norma_avg` is a normalization average using a specific pound when second position does not exists in document.
+
+The :code:`norma_avg`:
+- So if rambo occurs in 3 in movie synopsis in same document, we have 3 positions. In this case :code:`norma_avg` gets only first and second position to average.
+- Now if dance occur in 1 in music synopsis in same document, we have 1 position. In this case :code:`norma_avg` gets only first position and add a default pound in second position to average.
+
+The statistics available include:
+
+- :code: `raw_tp` -- the term position for a document. So if dance occurs 5 in music text in same document, this is 5 because of the term frequency.
+
 Finally a special stat exists for just counting the number of search terms. That stat is :code:`unique_terms_count`.
 
 ===========================
--- a/docs/feature-engineering.rst
+++ b/docs/feature-engineering.rst
@@ -48,7 +48,7 @@ The :code:`norma_avg`:
 
 The statistics available include:
 
-- :code: `raw_tp` -- the term position for a document. So if dance occurs 5 in music text in same document, this is 5 because of the term frequency.
+- :code:`raw_tp` -- the term position for a document. So if dance occurs 5 in music text in same document, this is 5 because of the term frequency.
 
 Finally a special stat exists for just counting the number of search terms. That stat is :code:`unique_terms_count`.
 
--- a/docs/feature-engineering.rst
+++ b/docs/feature-engineering.rst
@@ -44,8 +44,8 @@ For any of the cases, 0 will be returned if there isn't any occurrence of the te
 
 The statistics available include, e.g. using the query "dance monkey" we have:
 
-- :code:`min_raw_tp` -- return the minimal occurrence, i.e. the first one, of any term on the query. So if dance occurs at positions [2, 5 ,9], and monkey occurs at positions [1, 4] in a text in the same document, the minimal is 1.
-- :code:`max_raw_tp` -- return the maximum occurrence, i.e. the last one, of any term on the query. So if dance occurs at positions [2, 5 ,9] and monkey occurs at positions [1, 4] in a text in the same document, the minimal is 9.
+- :code:`min_raw_tp` -- return the minimum occurrence, i.e. the first one, of any term on the query. So if dance occurs at positions [2, 5 ,9], and monkey occurs at positions [1, 4] in a text in the same document, the minimum is 1.
+- :code:`max_raw_tp` -- return the maximum occurrence, i.e. the last one, of any term on the query. So if dance occurs at positions [2, 5 ,9] and monkey occurs at positions [1, 4] in a text in the same document, the maximum is 9.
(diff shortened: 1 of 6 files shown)


nathancday:
Thanks for the continued support you all! I'm sure others will find these features useful for their models too. Drop a comment when you think it's ready for review.

rtancman:
> Thanks for the continued support you all! I'm sure others will find these features useful for their models too. Drop a comment when you think it's ready for review.

@nathancday everything is ok to review ;)

src/main/java/com/o19s/es/explore/StatisticsHelper.java
```diff
@@ -84,4 +84,20 @@ public float getStdDev() {
 
         return (float) Math.sqrt(getVariance());
     }
+
+    public float getAvg() {
```
worleydl, line 88: Does this logic differ from calling getMean()?  We may want to add some comments explaining the difference.

src/main/java/com/o19s/es/explore/ExplorerQuery.java
```diff
@@ -214,6 +214,18 @@ public boolean isCacheable(LeafReaderContext ctx) {
...
             Query q = qb.build().rewrite(searcher.getIndexReader());
             return new ExplorerQuery.ExplorerWeight(this, searcher.createWeight(q, scoreMode, boost), type);
+        } else if (type.endsWith("_raw_tp")) {
+            // Rewrite this into a boolean query where we can inject our PostingsExplorerQuery
+            BooleanQuery.Builder qb = new BooleanQuery.Builder();
+            for (Term t : terms) {
+                qb.add(new BooleanClause(new PostingsExplorerQuery(t, PostingsExplorerQuery.Type.TP),
+                        BooleanClause.Occur.SHOULD));
+            }
+            // FIXME: completely refactor this class and stop accepting a random query but a list of terms directly
```
worleydl, line 224: We can leave out the duplicate fixme I think, I may make a helper call to clean up the duplicate logic too.

docs/feature-engineering.rst
```diff
@@ -37,6 +37,19 @@ The statistics available include:
 
 Putting the operation and the statistic together, you can see some examples. To get stddev of classic_idf, you would write :code:`stddev_classic_idf`. To get the minimum total term frequency, you'd write :code:`min_raw_ttf`.
 
+Term position statistics
+
+The :code:`type` parameter can be prepended with the operation to be performed across term position for the statistic :code:`max`, :code:`min`, :code:`avg`, and :code:`norma_avg`. The :code:`norma_avg` is a normalization average using a specific pound when second position does not exists in document.
```
worleydl, line 42: May need to clean this phrasing up a little bit.  I'm confused by "normalization average using a specific pound when second position does not exists in document.", also some small cleanup needed below.

src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java
```diff
@@ -189,4 +197,67 @@ public float getMaxScore(int upTo) throws IOException {
             return Float.POSITIVE_INFINITY;
         }
     }
+
+    static class TPScorer extends PostingsExplorerScorer {
+        TPScorer(Weight weight, PostingsEnum postingsEnum) {
+            super(weight, postingsEnum);
+        }
+        final float NORMALIZED_TP = 5000.0f;
```
worleydl, line 205: Slightly nervous about this constant.  Will need to review the logic after gaining a better understanding of what normalized average is accomplishing.

rtancman: This is a default weight to penalize documents without term second position.

rtancman: @MaquisW may explain it the better way to @worleydl why we created this feature.

src/main/java/com/o19s/es/explore/StatisticsHelper.java
rtancman: I dind`t see `getMean` 😔. We will fix it

src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java
MaquisW: That's exactly what @rtancman explained. In this case, `norma_avg` is computed only with the first two occurrences of any query term, the idea is that this feature demonstrates how far, on average, the first two occurrences of any query term are in the document. With that in mind, it is possible that documents that have only one occurrence may have an advantage in calculating this average, so we simulate a fake second occurrence. Ideally, this fake occurrence should penalize the calculation of the average of the first 2 occurrence positions in this document, considering that none of the terms of the query occurred a second time, which must be bad for the document's score. One way to penalize is to simulate that the fake occurrence happens at the end of the document, so `NORMALIZED_TP` (which needs a better name, by the way) could have a value like `size_of_the_document + 1`, but small documents could continue to have an advantage (and we also had trouble getting the document size at this point), so we assume a high value like 5000. We are still experimenting and studying the engineering possibilities of this feature, and we saw that all these calculations still have some issues. We appreciate the feedback, soon we will update the PR with more information and clearer documentation. Thank you!

nathancday: Thanks for the explanation. I can see the value of the first N occurrences feature, and I also agree there is experimentation to be done about how to normalize for document length and what to do with <N occurrences. 

Because this feature is still being developed, would you be okay if we only include `min`, `max` and `avg` in the PR for now?

We just want to make sure these the plugin supports features that a broadly useful for a variety of Elasticsearch users/cases.

docs/feature-engineering.rst
```diff
@@ -37,6 +37,17 @@ The statistics available include:
...
 Putting the operation and the statistic together, you can see some examples. To get stddev of classic_idf, you would write :code:`stddev_classic_idf`. To get the minimum total term frequency, you'd write :code:`min_raw_ttf`.
 
+Term position statistics
+
+The :code:`type` parameter can be prepended with the operation to be performed across term position for the statistic :code:`min`, :code:`max` and :code:`avg`.
+For any of the cases, 0 will be returned if there isn't any occurrence of the terms in the document.
+
+The statistics available include, e.g. using the query "dance monkey" we have:
+
+- :code:`min_raw_tp` -- return the minimal occurrence, i.e. the first one, of any term on the query. So if dance occurs at positions [2, 5 ,9], and monkey occurs at positions [1, 4] in a text in the same document, the minimal is 1.
```
worleydl, line 47: It doesn't seem like I can edit your PR directly, can you change these lines to look like the following:

```
- :code:`min_raw_tp` -- return the minimum occurrence, i.e. the first one, of any term on the query. So if dance occurs at positions [2, 5 ,9], and monkey occurs at positions [1, 4] in a text in the same document, the minimum is 1.
 - :code:`max_raw_tp` -- return the maximum occurrence, i.e. the last one, of any term on the query. So if dance occurs at positions [2, 5 ,9] and monkey occurs at positions [1, 4] in a text in the same document, the maximum is 9.
```

Changes: `s/minimal/minimum`, and the description for max was using min in its definition.

If we can get this updated I think we're ready to merge.  Thanks for the updates!

PR #292, opened 2020-04-04
Fixes #286 - Add feature normalization as part of model definition

### LINKED ISSUE #286: Feature Normalization
I was recently using RankSVM with Solr, and realized through that process how crucial feature normalization is with SVMs. I wondered about the Elastic LTR plugin, and realized that we probably need to add optional, per-feature normalization.

In a perfect world we could 'log' normalized features. Then train using the same feature normalization that is used during model evaluation in the plugin.

However, that's probably not possible as most normalization methods are either min/max (what's the min / max value for this feature) or are based on mean / std deviation of the feature. So you need to see the full range of possible feature values when training. This IS feasible when you're training over the whole set. 

Still these are per-feature characteristics, so it would seem that either the model or the features set needs to associate a feature with a normalizer... and that this could be updated via a CRUD API

softwaredoug:
For issue #286 

It's a bit more extensive of a change that I am used to, I wanted to get @nomoa and other's perspective on feature normalization addition. There is a lot of remaining work / cleanup to do, but I wanted to get consensus on the approach from anyone interested.

1. Adds a 'feature_normalization' attribute of a model definition. With two normalizers - min/max and standard. Here's an example of the syntax:

```
{
	"name": "my_model",
	"feature_set": "foo",
	"model": {
		"type": "model/dummy",
		"definition": "(any model den)",
		"feature_normalizers": {
			"feature_2": {
				"min_max": {
					"minimum": -1.0,
					"maximum": 1.25
				}
			}
		}
	}
}
```

2. At model execution time, we wrap a feature vector with one that normalizes prior to retrieving the values. The model than uses normalized values instead of raw ones. So there is a wrapping ranker for feature normalization

3. I changed logging to only work with a NullRanker, as I couldn't see where we would use another kind of ranker. As logging should never log normalized feature values, I wanted to ensure it was always using the NullRanker purpose built for logging.

Some things to notice
- The parsing code is the most complex / annoying. Please note and give feedback on that code
- There is an existing Normalizer interface, that likely we ought to merge with FeatureNormalizer (and we may wish to distinguish between normalizing model output vs normalizing features)

TODO:

- [  ] Test model inference with normalized features
- [  ] Integration testing
- [  ] Code cleanup
- [  ] Documentation

Also, please send PRs or make changes to this branch!


softwaredoug:
This is a bit less WIP now, so removed WIP. Though I recognize we need documentation, integration tests, and a pass through the code to make sure it's clean

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+
+            BiConsumer<FeatureNormConsumer, MinMaxFeatureNormDefinition> setMinMax = (s, v) -> {
+                v.setFeatureName(featureName);
+                s.setFtrNormDefn(v);
+            };
+
+            parser.declareObject(setStd, StandardFeatureNormDefinition.PARSER, STANDARD);
+            parser.declareObject(setMinMax, MinMaxFeatureNormDefinition.PARSER, MIN_MAX);
+
+            FeatureNormConsumer parsedNorm  = parser.parse(p, c);
```
nomoa, line 74: Instead of Void for the context you could switch to String here and pass featureName this there's no need to set featureName after the fact and you don't need the two BiConsumer above as you may call FeatureNormConsumer::setFtrNormDefn directly I believe

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+
+        FeatureNormConsumer() {
+        }
+
+        public FeatureNormDefinition getFtrNormDefn() {
+            return this.ftrNormDefn;
+        }
+
+        public void setFtrNormDefn(FeatureNormDefinition ftrNormDefn) {
+            this.ftrNormDefn = ftrNormDefn;
```
nomoa, line 94: should prevent setting the normalizer twice by throwing IllegalArgumentException if this.ftrNormDefn is already set?

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+            FeatureNormDefinition norm = this.createFromStreamInput(input);
+            this.featureNormalizers.put(norm.featureName(), norm);
+        }
+    }
+
+    public Normalizer getNormalizer(String featureName) {
+        return this.featureNormalizers.get(featureName).createFeatureNorm();
+    }
+
+    public static final String TYPE = "feature_normalizers";
```
nomoa, line 118: is this TYPE used?

src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
```diff
@@ -228,29 +252,37 @@ void setRankingModel(LtrModelDefinition rankingModel) {
...
             modelAsString = in.readBoolean();
+            this.featureNormalizerSet = new StoredFeatureNormalizers(in);
         }
 
         @Override
         public void writeTo(StreamOutput out) throws IOException {
             out.writeString(type);
             out.writeString(definition);
             out.writeBoolean(modelAsString);
+            this.featureNormalizerSet.writeTo(out);
```
nomoa, line 285: same problem as when reading

src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
```diff
@@ -228,29 +252,37 @@ void setRankingModel(LtrModelDefinition rankingModel) {
...
             this.definition = definition;
             this.modelAsString = modelAsString;
+            this.featureNormalizerSet = new StoredFeatureNormalizers();
         }
 
         public LtrModelDefinition(StreamInput in) throws IOException {
             type = in.readString();
             definition = in.readString();
             modelAsString = in.readBoolean();
+            this.featureNormalizerSet = new StoredFeatureNormalizers(in);
```
nomoa, line 277: this is a breaking change

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+            return this.ftrNormDefn;
+        }
+
+        public void setFtrNormDefn(FeatureNormDefinition ftrNormDefn) {
+            this.ftrNormDefn = ftrNormDefn;
+        }
+    }
+
+
+    private Map<String, FeatureNormDefinition> featureNormalizers;
```
nomoa, line 99: could be final?

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return this.featureNormalizers.hashCode();
+    }
+
+
+    private  FeatureNormDefinition createFromStreamInput(StreamInput input) throws IOException {
+        Type normType = Type.readFromStream(input);
```
nomoa, line 179: you should perhaps use the registry:
In the plugin class in getNamedWritables you could add the entries:
```java
new Entry(FeatureNormDefinition.class, MinMaxFeatureNormDefinition.TYPE, MinMaxFeatureNormDefinition::new)
```

This way you just have to call:
```List<FeatureNormDefinition> featureNormalizers = input.readNamedWritableList(FeatureNormDefinition.class)```
in the constructor of this class.

File: src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java [hunk 8]
```diff
@@ -0,0 +1,197 @@
...
+        } else if (normType == Type.MIN_MAX) {
+            return new MinMaxFeatureNormDefinition(input);
+        }
+        // note the Type constructor throws on this condition as well
+        throw new ElasticsearchException("unknown normalizer type during deserialization");
+    }
+
+    public void writeTo(StreamOutput output) throws IOException {
+        output.writeInt(this.featureNormalizers.size());
+        for (Map.Entry<String, FeatureNormDefinition> featureNormEntry: this.featureNormalizers.entrySet()) {
```
nomoa, line 191: same here if you have the implementation declared in the registry simply calling `output.writeNamedWriteableList(new ArrayList(this.featureNormalizers.values()))`

src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
```diff
@@ -79,6 +86,7 @@ public StoredLtrModel(StreamInput input) throws IOException {
         rankingModelType = input.readString();
         rankingModel = input.readString();
         modelAsString = input.readBoolean();
+        this.parsedFtrNorms = new StoredFeatureNormalizers(input);
```
nomoa, line 89: this is a breaking change, and will break while doing a rolling restart to ship this version of the plugin.
The last time we had to do this we guarded against the next version this feature will be shipped to, e.g. if this feature is shipped with v7.6.3 writing the normalizers must be guarded with:

        if (input.getVersion().onOrAfter(Version.V_7_6_3)) {
             new StoredFeatureNormalizers(input);
        }

src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
```diff
@@ -88,6 +96,7 @@ public void writeTo(StreamOutput out) throws IOException {
         out.writeString(rankingModelType);
         out.writeString(rankingModel);
         out.writeBoolean(modelAsString);
+        parsedFtrNorms.writeTo(out);
```
nomoa, line 99: breaking change

src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
```diff
@@ -262,6 +294,10 @@ public String getType() {
             return type;
         }
 
+        public void setNamedFeatureNormalizers(List<FeatureNormDefinition> featureNormalizers) {
+            this.featureNormalizerSet.addFeatureNormalizers(featureNormalizers);
```
nomoa, line 298: instead of adding I that it should set a this.featureNormalizerSet = new StoredFeatureNormalizers(featureNormalizers);

src/main/java/com/o19s/es/ltr/ranker/normalizer/FeatureNormalizingRanker.java
```diff
@@ -0,0 +1,72 @@
...
+import com.o19s.es.ltr.feature.FeatureNormalizerSet;
+import com.o19s.es.ltr.ranker.LtrRanker;
+import org.apache.lucene.util.Accountable;
+import org.apache.lucene.util.RamUsageEstimator;
+
+/**
+ * An LtrRanker that applies supplied normalizers to features prior
+ * to scoring
+ */
+public class FeatureNormalizingRanker implements LtrRanker, Accountable {
```
nomoa, line 12: Another strategy is to wrap feature queries with a NormalizingQueryWrapper that accepts the featureQuery and the normalizer. This would be built in the RankerQuery::createWeight when looping over the queries, the normalizers could be passed as a new field in LtrQueryContext.

The benefits I could see are:
- no need to loop again over the features
- we would not lose the explanation

not sure if it's worth effort redesign what you already wrote. If you are interested I think I may some code ready (or close to be ready) for this NormalizingQueryWrapper query class.

src/main/java/com/o19s/es/ltr/ranker/normalizer/MinMaxFeatureNormalizer.java
```diff
@@ -0,0 +1,53 @@
...
+                                             " must be smaller than than maximum: " +
+                                              Double.toString(maximum));
+        }
+        this.minimum = minimum;
+        this.maximum = maximum;
+    }
+
+    @Override
+    public float normalize(float value) {
+        return  (value - minimum) / (maximum - minimum);
```
nomoa, line 29: should this be guarded to avoid negative values?

src/main/java/com/o19s/es/ltr/ranker/normalizer/MinMaxFeatureNormalizer.java
```diff
@@ -0,0 +1,53 @@
...
+ * See
+ * https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html
+ */
+public class MinMaxFeatureNormalizer implements Normalizer  {
+    float maximum;
+    float minimum;
+
+    public MinMaxFeatureNormalizer(float minimum, float maximum) {
+        if (minimum >= maximum) {
+            throw new ElasticsearchException("Minimum " + Double.toString(minimum) +
```
nomoa, line 19: we should use IllegalArgumentException I think

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+        builder.startObject(); // begin feature norms
+        for (Map.Entry<String, FeatureNormDefinition> ftrNormDefEntry: featureNormalizers.entrySet()) {
+            builder.field(ftrNormDefEntry.getKey());
+            ftrNormDefEntry.getValue().toXContent(builder, params);
+        }
+        builder.endObject(); // feature norms
+        return builder;
+    }
+
+    public void addFeatureNormalizers(List<FeatureNormDefinition> ftrNormDefs) {
```
nomoa, line 131: I think it'd make more sense to have a constructor that accepts a List<FeatureNormDefinition> ftrNormDefs instead of a add method
It seems that whenever we call this method we are actually constructing a new StoredFeatureNormalizers

src/main/java/com/o19s/es/ltr/feature/store/FeatureNormDefinition.java
```diff
@@ -0,0 +1,28 @@
+package com.o19s.es.ltr.feature.store;
+
+import com.o19s.es.ltr.ranker.normalizer.Normalizer;
+
+/**
+ * Parsed feature norm from model definition
+ */
+public interface FeatureNormDefinition extends StorableElement {
```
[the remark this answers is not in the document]
softwaredoug: I do need a FeatureNormDefinition to be able to serialize / deserialize to and from xContent and a StreamInput/Binary. Is this something you would recommend just putting directly in the interface?

src/main/java/com/o19s/es/ltr/feature/store/MinMaxFeatureNormDefinition.java
```diff
@@ -0,0 +1,129 @@
...
+
+    MinMaxFeatureNormDefinition(StreamInput input) throws IOException {
+        this.featureName = input.readString();
+        this.minimum = input.readFloat();
+        this.maximum = input.readFloat();
+    }
+
+    public MinMaxFeatureNormDefinition() {
+        this.maximum = Float.MAX_VALUE;
+        this.minimum = Float.MIN_VALUE;
```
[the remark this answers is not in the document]
softwaredoug: Hmm good call out. I didn't notice that! I need to think about this, and will leave another comment on the PR about whether it should be a -MAX_VALUE or not.

src/main/java/com/o19s/es/ltr/ranker/normalizer/FeatureNormalizingRanker.java
softwaredoug: I think that's a fine approach too... If it's small enough, want to give it a try on this branch?

softwaredoug:
Cool thanks, did all the low-hanging fruit review comments.

**On negative scores**: I think our policy probably needs to continue to be to let the user figure this out. I hesitate to try to figure out someone's model they submitted and try to patch it or even reject it prematurely if, in the end, it behaves OK. But open to suggestions, this is a tricky issue for sure...

**On binary compatibility**, that should be fixable I think. I can use `available` on StreamInput to check if any bytes are available when constructing FeatureNormalizers class from StreamInput... and if none are available, construct empty FeatureNormalizers. I'll submit something, hopefully with a test... 

I also have an integration test that's failing, where I appear to not be able to retrieve the feature normalizers during CRUD operations. Hopefully I'll have time to debug this this week, but let me know if anything pops into your head...

src/main/java/com/o19s/es/ltr/ranker/normalizer/MinMaxFeatureNormalizer.java
softwaredoug: perhaps for min / max it should, as we should not have features with negative values logged from ES

softwaredoug:
Ah integ test was just a minor flub... fixed!

src/main/java/com/o19s/es/ltr/feature/store/StoredLtrModel.java
softwaredoug: I changed this to `after(V_7_6_1)` so I suppose this will be active when we update to 7.6.2

src/main/java/com/o19s/es/ltr/feature/store/StoredFeatureNormalizers.java
```diff
@@ -0,0 +1,197 @@
...
+            throw new ElasticsearchException("unknown normalizer type during serialization [" + ord + "]");
+        }
+
+        @Override
+        public void writeTo(StreamOutput out) throws IOException {
+            out.writeVInt(this.ordinal());
+        }
+    }
+
+    public static final ObjectParser.NamedObjectParser<FeatureNormDefinition, Void> PARSER;
```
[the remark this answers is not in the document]
softwaredoug: Yes, this actually worked - thanks! Much more concise

softwaredoug:
I think I have resolved all the comments, I think I pretty much took all your feedback

- On binary compatibility, I probably need to change this to 7.6.3 whenever 7.6.3 comes out :-/ 
- Note the change from the custom Ranker to a Feature Set that wraps a feature with a query (always feel free to double check my equals implementations :) )
- Many other changes, etc

softwaredoug:
There is an issue with wrapping queries to implement feature normalization. It doesn't deal with normalizing scores from non-matching documents. When we set missing values to 0, these 0s also need to be normalized.

nomoa:
> The problem is a wrapping scorer doesn't score documents it doesn't match. So you fail to normalize non matching scores (ie normalize 0). When the score is 0, the resulting normalized feature value would probably be non-zero. Yet when we miss features now the feature value ends up a non-normalized 0, creating incorrect model evaluation.

If if the `missing value` needs to be normalized then indeed wrapping the query may not give the opportunity to run the normalizer on the value 0 if it's assumed to be the missing value.
Ideally we should be computing this `normalized missing value` once per feature and consider it a constant. One place to do so is inside `com.o19s.es.ltr.ranker.LtrRanker#newFeatureVector` implementations and instead of filling the vector with 0s it'd be filled with the normalized values that could have been pre-computed at model creation and simply copied over.
In the end this will make the Ranker implementation aware the feature normalizers.
I haven't dug into the details and it may not be worthwhile nor practical to implement and thus your first approach should perhaps be reconsidered. Sorry that this led to this outcome...

softwaredoug:
> > The problem is a wrapping scorer doesn't score documents it doesn't match. So you fail to normalize non matching scores (ie normalize 0). When the score is 0, the resulting normalized feature value would probably be non-zero. Yet when we miss features now the feature value ends up a non-normalized 0, creating incorrect model evaluation.
> 
> If if the `missing value` needs to be normalized then indeed wrapping the query may not give the opportunity to run the normalizer on the value 0 if it's assumed to be the missing value.
> Ideally we should be computing this `normalized missing value` once per feature and consider it a constant. One place to do so is inside `com.o19s.es.ltr.ranker.LtrRanker#newFeatureVector` implementations and instead of filling the vector with 0s it'd be filled with the normalized values that could have been pre-computed at model creation and simply copied over.
> In the end this will make the Ranker implementation aware the feature normalizers.
> I haven't dug into the details and it may not be worthwhile nor practical to implement and thus your first approach should perhaps be reconsidered. Sorry that this led to this outcome...

No worries! I think it has been useful to explore the alternatives.

Aside from the issue of wanting a 7.6.3 version for binary compatibility, this PR I think is ready. Please feel free to do another review!

src/main/java/com/o19s/es/ltr/ranker/normalizer/FeatureNormalizingRanker.java
```diff
@@ -0,0 +1,71 @@
...
+    public int hashCode() {
+        return this.wrapped.hashCode() +
+                (31 * this.ftrNorms.hashCode());
+    }
+
+    @Override
+    public long ramBytesUsed() {
+        Accountable accountable = (Accountable)this.wrapped;
+        if (accountable != null) {
+            return RamUsageEstimator.NUM_BYTES_OBJECT_HEADER + accountable.ramBytesUsed();
```
nomoa, line 66: storing a static BASE_RAM_USED = RamUsageEstimator.shallowSizeOfInstance(FeatureNormalizingRanker.class) and using it inplace of RamUsageEstimator.NUM_BYTES_OBJECT_HEADER might be more accurate.

We also lack the size of the normalizers themselves, I have no clue if it's worthwhile to implement or just roughly estimate its size like ftrNorms.size() * randomGuess.

For now ram estimation is only useful for cache eviction strategies

src/main/java/com/o19s/es/ltr/ranker/normalizer/NoOpNormalizer.java
```diff
@@ -0,0 +1,20 @@
+package com.o19s.es.ltr.ranker.normalizer;
+
+public class NoOpNormalizer implements Normalizer {
```
nomoa, line 3: I think you might be able to just remove this class now

src/main/java/com/o19s/es/ltr/ranker/normalizer/Normalizer.java
```diff
@@ -1,8 +1,14 @@
...
 
+import org.apache.lucene.search.Explanation;
+
 /**
  * Interface to normalize the resulting score of a model
  */
 public interface Normalizer {
     float normalize(float val);
+
+    default Explanation explain(Explanation wrappedQueryExplain) {
```
nomoa, line 11: perhaps unneeded now

src/main/java/com/o19s/es/ltr/feature/store/FeatureNormDefinition.java
```diff
@@ -0,0 +1,41 @@
...
+    StoredFeatureNormalizers.Type normType();
+
+    /**
+     * Serialize to a StreamOutput
+     * @param out
+     *  where the ftr norm defn will be serialized
+     * @throws
+     *  IOException if issues occur with writing to output
+     */
+    void writeTo(StreamOutput out) throws IOException;
```
nomoa, line 40: you should probably extend org.elasticsearch.common.io.stream.Writeable as well so that you don't have to re-declare this method.

softwaredoug:
Just a note before merging, this PR needs some docs

softwaredoug:
Docs done, going to merge after changing the feature flags on binary decoding / encoding to 7.7

src/test/java/com/o19s/es/ltr/query/LtrQueryTests.java
```diff
@@ -301,10 +325,9 @@ public void checkModelWithFeatures(List<PrebuiltFeature> features, int[] modelFe
         String modelAsStr = ranker.model();
         RankerFactory rankerFactory = new RankerFactory();
         Ranker rankerAgain = rankerFactory.loadRankerFromString(modelAsStr);
-        float[] scoresAgain = new float[] {(float)rankerAgain.eval(rl.get(0)), (float)rankerAgain.eval(rl.get(1)),
-                (float)rankerAgain.eval(rl.get(2)), (float)rankerAgain.eval(rl.get(3))};
+        float[] scoresAgain = {(float)ranker.eval(rl.get(0)), (float)ranker.eval(rl.get(1)),
```
sstults, line 328: @softwaredoug It's been a while since you made this change, but do you recall why you're referencing `ranker` here instead of `rankerAgain`?

Issue #293, opened 2020-04-07
Add parameters in term position feature


MaquisW:
Relating to the new feature added in #288, we would like to add the possibility to specify a position we desire to obtain the term statistics.

worleydl:
Resolved by TermStatQuery: https://elasticsearch-learning-to-rank.readthedocs.io/en/latest/advanced-functionality.html#termstat-query

PR #294, opened 2020-04-07
Add parameters in term position

MaquisW:
Refer #293


softwaredoug:
For the time being, going to hold off as @worleydl sees if we can do this in a derived expression or some other script...

softwaredoug:
Thanks @MaquisW - appreciate the PR. As @worleydl is working on scriptable term stats to address this issue, I am going to go ahead and close the PR. Appreciate you helping to trigger discussion around this area.

Issue #296, opened 2020-04-08
Add classic idf multiplied by tf in match_explorer


rtancman:
We want to add new feature `idf_x_tf` in `match_explorer`. This feature is referenced in [Microsoft's LETOR](https://www.microsoft.com/en-us/research/project/letor-learning-rank-information-retrieval/).

-------------
_LETOR is a package of benchmark data sets for research on LEarning TO Rank, which contains standard features, relevance judgments, data partitioning, evaluation tools, and several baselines._
> https://arxiv.org/ftp/arxiv/papers/1306/1306.2597.pdf
> Tao Qin and Tie-Yan Liu. Introducing LETOR 4.0 Datasets, arXiv preprint arXiv:1306.2597.

nomoa:
Not entirely convinced that the match_explorer is the right place for this kind of features. This query was designed to help surfacing low level features stored in lucene. I believe that the `derived_expression` feature of the plugin can be used for that. There's also an experimental feature we added in another [plugin](https://github.com/wikimedia/search-extra/blob/master/docs/simswitcher.md) that allows to experiment with different similarities (without having to close/reopen an index). Note that elastic also provides a nice way to create custom similarities using painless script (https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-similarity-context.html).

softwaredoug:
@worleydl gonna see if this is doable in a derived expression... 

I agree with the general principal if there was a way to script these various things (both the positional info & the raw idf tf math) then we would prefer that. It would give a lot of freedom to experiment with many / different types of features.

rtancman:
@nomoa @softwaredoug I saw the [similarity module](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-similarity.html#index-modules-similarity) and I have some questions. We want to use BM25 scores in all fields but in rescore, we need to create features using TF and IDF. Is there a way to create this feature without change the BM25 scores by fields? 

I will test that with `weight_script` to check the score.

worleydl:
@rtancman here's an example of creating custom formulas using the raw statistics via a script feature: https://gist.github.com/worleydl/4b3c7cf150185459aa538bc7f4f63a52

Regarding the BM25 question, you should be able to setup features that utilize the defined similarity for your index in conjunction with features that are using raw statistics from the match explorer.  The only caveat I see right now is that any statistics you wish to use in your custom formula need to be included in the feature set.

[...]

PR #297, opened 2020-04-08
Add classic idf multiplied by tf in match_explorer

rtancman:
refer: #296


softwaredoug:
For the time being, going to hold off as @worleydl sees if we can do this in a derived expression or some other script...

softwaredoug:
Thanks @rtancman as @worleydl is working on scriptable term stats features, I think we're going to close it. Appreciate triggering this discussion though, it's been very valuable. It likely will take a bit of time before it's done & dusted so it's understandable if temporarily you need to create your own plugin for these kinds of features

PR #306, opened 2020-05-14
Upgrade to Elasticsearch 7.7

softwaredoug:
Set `.ltrstore` index to hidden

Gradle Wrapper Upgrade to 6.4

REST Handler needs to declare routes method instead of creating routes in their constructor. See [this commit](
https://github.com/elastic/elasticsearch/commit/627d853f18c65a3ae7d23dd455db8b061eaad104) for Elastic's similar change.

TermQueryBuilder now rewritten when field doesn't exist. Many tests assumed TermQueryBuilder was never rewritten. The test in LTR Query Builder was changed to MatchQueryBuilder with comment that we should not assume this would hold true in the future

ExplorerQueryBuilder similarly uses a term query builder for testing. Specifically, a term query builder for any empty field (as query builder tests don't index documents). In such cases TermQueryBuilder is rewritten to MatchNoDocs Query. So testing was added to confirm that in this case, ExplorerQueryBuilder does indeed rewrite correctly.


nomoa:
For the index naming issue perhaps that setting `index.hidden: true` will silence this warning for now, in com.o19s.es.ltr.feature.store.index.IndexFeatureStore#storeIndexSettings by setting: 
`.put(IndexMetaData.SETTING_INDEX_HIDDEN, true)`

If this works I believe that this can be changed on existing indices by tweaking the index setting `index.hidden` to true.

softwaredoug:
For some reason tonight, Travis is refusing to install OpenJDK14 or OracleJDK14. See [this post](https://travis-ci.community/t/installing-java-14-fails-with-travis-setup-java-command-not-found/8626). Grrr....

softwaredoug:
All [Hello LTR](http://github.com/o19s/hello-ltr) notebooks run successfully for ES 7.7 and this branch.

Issue #304, opened 2020-05-06
Add setting for enable/disable plugin


jngz-es:
I'd like to add a cluster setting for enable/disable plugin's functionalities.

nathancday:
Can you point me to another plugin where this is done? Trying to figure out a common use case, right now it feels similar to node roles.

jngz-es:
It's similar with https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/105.

nathancday:
Thanks for sharing that ref. If this becomes desired by more folks we may consider developing it. In the mean time, if you feel like making a PR based on the AD work, that would be lovely. We are always happy to have new project contributors!

PR #309, opened 2020-05-30
add settings for disable/enable LTR plugin

jngz-es:
#304

### DIFF
--- a/src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java
+++ b/src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java
@@ -16,6 +16,7 @@
 
 package com.o19s.es.explore;
 
+import com.o19s.es.ltr.LTRSettings;
 import com.o19s.es.ltr.utils.CheckedBiFunction;
 import org.apache.lucene.index.IndexReaderContext;
 import org.apache.lucene.index.LeafReaderContext;
@@ -78,6 +79,10 @@ public class PostingsExplorerQuery extends Query {
     @Override
     public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost)
             throws IOException {
+        if (!LTRSettings.isLTRPluginEnabled()) {
+            throw new IllegalStateException("LTR plugin is disabled. To enable update ltr.plugin.enabled to true");
+        }
+
         IndexReaderContext context = searcher.getTopReaderContext();
         assert scoreMode.needsScores() : "Should not be used in filtering mode";
         return new PostingsExplorerWeight(this, this.term, TermStates.build(context, this.term,
--- a/build.gradle
+++ b/build.gradle
@@ -78,3 +78,8 @@ thirdPartyAudit.enabled = false
 
 // Uncomment this to skip license header checks
 licenseHeaders.enabled = false
+
+// We don't need to follow ES testing naming conventions.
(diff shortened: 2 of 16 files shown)


nathancday:
Awesome, thanks for donating this! Give me some time to get this reviewed and I'll be back. Thanks again for building this out for the plugin.

worleydl:
Will review this week.

jngz-es:
> Will review this week.

Thanks.

worleydl:
I have some edits I'm working on for this.  @nomoa what do you think of this concept as a whole?  I haven't seen enable/disable settings in other plugins I wonder if this is better served elsewhere somehow instead of having code scattered around to check the setting within the plugin.

@jngz-es is this a common pattern you've seen in other plugins?

src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java
```diff
@@ -78,6 +79,10 @@ public int hashCode() {
     @Override
     public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost)
             throws IOException {
+        if (!LTRSettings.isLTRPluginEnabled()) {
```
worleydl, line 82: Possible edit is making this a "checkAndThrow" type of call as to not have the exception throw call be duplicated.

jngz-es:
> I have some edits I'm working on for this. @nomoa what do you think of this concept as a whole? I haven't seen enable/disable settings in other plugins I wonder if this is better served elsewhere somehow instead of having code scattered around to check the setting within the plugin.
> 
> @jngz-es is this a common pattern you've seen in other plugins?

@worleydl https://github.com/opendistro-for-elasticsearch/anomaly-detection, https://github.com/opendistro-for-elasticsearch/k-NN.

src/main/java/com/o19s/es/explore/PostingsExplorerQuery.java
jngz-es: Good point.

nomoa:
> I have some edits I'm working on for this. @nomoa what do you think of this concept as a whole? I haven't seen enable/disable settings in other plugins I wonder if this is better served elsewhere somehow instead of having code scattered around to check the setting within the plugin.
> 

I haven't seen this pattern either and have the same concerns, a central point of the plugin is the feature store so perhaps guarding on this condition could made there and would be a stronger guarantee that the plugin can't be used (except the couple queries we have that do rely on it LtrQuery/MatchExplorer).

jngz-es:
@worleydl @nomoa I got your points. Do you have any suggestion on this feature? I mean like where to have a central control?

nomoa:
@jngz-es sorry for the late reply.

I think adding the check in:
- com.o19s.es.ltr.feature.store.index.Caches#cacheLoad
- com.o19s.es.ltr.ranker.parser.LtrRankerParserFactory#getParser
will cover the most important and most costly operations of the plugin and will make the plugin completely useless.

If the goal is to disable absolutely everything (e.f. feature management) then you'll have to add the checks to:
- all rest endpoints in the package `com.o19s.es.ltr.rest`
- com.o19s.es.explore.ExplorerQueryBuilder#doToQuery (this is the sole query usable without the infrastructure of the plugin, i.e. it could be moved into a separate plugin).

With this I think all features of the plugin should be disabled.

PR #313, opened 2020-06-15
TermStatQuery and ScriptFeature term stats

worleydl:
This PR addresses #296 and partially #293 

### TermStatQuery
The TermStatQuery is an attempt to clean up some of the code inside of the ExplorerQuery.  Instead of accepting a query and a stat type it accepts a raw specification of terms and allows for a custom lucene expression to evaluate the term statistics for each document.

### ScriptFeature Term Stats
Since the concept seemed doable I've added some functionality to the ScriptFeature so it will inject TermStats into the script params if the user requests it.  This provides access to the actual collections of term statistics so it should open up lots of avenues for feature experimentation.

## TODO:
- [x] Experiment with query visitor behaviors and possibly add support
- [x] Build out functionality for different formats of passing terms in
- [x] Documentation


worleydl:
Hey @nomoa I'll be picking this back up this weekend.  If you get a chance I'm curious if you have any thoughts about the viability of "injecting" the term stats into the ScriptFeature.  Also any thoughts you have about specifying terms would be appreciated. Thanks!

src/main/java/com/o19s/es/termstat/TermStatQuery.java
```diff
@@ -0,0 +1,122 @@
...
+            if (newDoc == doc) {
+                return Explanation
+                        .match(scorer.score(), "weight(" + this.expression + " in doc " + newDoc + ")");
+            }
+            return Explanation.noMatch("no matching term");
+        }
+
+        @Override
+        public Scorer scorer(LeafReaderContext context) throws IOException {
+            Expression compiledExpression = (Expression) Scripting.compile(this.expression);
```
nomoa, line 113: you might perhaps want to compile this earlier (in TermStatQueryBuilder and pass Expression compiledExpression in the constructor of this query) so that it's built once per query as opposed to once per segment.

src/main/java/com/o19s/es/termstat/TermStatScorer.java
```diff
@@ -0,0 +1,155 @@
...
+
+        for(int i = 0; i < tf_stats.getSize(); i++) {
+            // Update the term stat dictionary for the current term
+            termStatDict.put("df", df_stats.getData().get(i));
+            termStatDict.put("idf", idf_stats.getData().get(i));
+            termStatDict.put("tf", tf_stats.getData().get(i));
+            termStatDict.put("tp", tp_stats.getData().get(i));
+            termStatDict.put("ttf", ttf_stats.getData().get(i));
+
+            DoubleValuesSource dvSrc = compiledExpression.getDoubleValuesSource(bindings);
```
nomoa, line 141: Ideally this should be called once when creating the scorer, this will drastically improve the perf I believe since here you rescan the term enum on every doc.
Might not be trivial to achieve but I think it might worth it (perhaps in a separate subsequent PR?)

src/main/java/com/o19s/es/termstat/TermStatQueryBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+
+    private static final ParseField EXPR_NAME = new ParseField("expr");
+    private static final ParseField AGGR_NAME = new ParseField("aggr");
+    private static final ParseField POS_AGGR_NAME = new ParseField("pos_aggr");
+    private static final ParseField TERMS_NAME = new ParseField("terms");
+
+    private String expr;
+    private AggrType aggr;
+    private AggrType pos_aggr;
+    private Set<Term> terms;
```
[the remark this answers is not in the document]
worleydl: I've updated to an array of terms and fields with an optional analyzer param. By default it uses the search analyzer configured for the field.

src/main/java/com/o19s/es/termstat/TermStatQuery.java
worleydl: Done.

src/main/java/com/o19s/es/termstat/TermStatScorer.java
worleydl: I've played with this a bit but haven't figured out the magic to get it working yet.

PR #314, opened 2020-06-16
Update to Elasticsearch 7.7.1

thePanz:
Moved to gradle.properties, as seen in: https://github.com/spinscale/elasticsearch-ingest-opennlp


nathancday:
Thank you!

PR #315, opened 2020-06-23
Update to ES 7.8.0

thePanz:
*Help needed*

- [x] Fix case-change in IndexMetadata and MetadataCreateIndexService class names
- [x] Fix overrided method LtrQueryParserPlugin::createComponents()
- [x] Fix signature change for SearchHit() constructor in LoggingFetchSubPhaseTests.java line 175
- [x] Fix failing tests, example: https://pastebin.com/5Yneh9LV @nomoa !
- [x] Fix failing tests, and exception: https://pastebin.com/LqVd9evd thanks @nomoa !

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
 ltrVersion = 1.3.0
 elasticsearchVersion = 7.8.0
-luceneVersion = 8.5.0
+luceneVersion = 8.5.1
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,3 @@
 ltrVersion = 1.3.0
-elasticsearchVersion = 7.7.1
+elasticsearchVersion = 7.8.0
+luceneVersion = 8.5.0
--- a/src/test/java/com/o19s/es/ltr/query/StoredLtrQueryBuilderTests.java
+++ b/src/test/java/com/o19s/es/ltr/query/StoredLtrQueryBuilderTests.java
@@ -46,6 +46,7 @@ import org.elasticsearch.index.query.functionscore.FieldValueFactorFunctionBuild
 import org.elasticsearch.index.query.functionscore.FunctionScoreQueryBuilder;
 import org.elasticsearch.plugins.Plugin;
 import org.elasticsearch.test.AbstractQueryTestCase;
+import org.elasticsearch.test.TestGeoShapeFieldMapperPlugin;
 import org.junit.Before;
 
 import java.io.IOException;
@@ -65,8 +66,9 @@ import static org.hamcrest.CoreMatchers.instanceOf;
 public class StoredLtrQueryBuilderTests extends AbstractQueryTestCase<StoredLtrQueryBuilder> {
     private static final MemStore store = new MemStore();
 
+    // TODO: Remove the TestGeoShapeFieldMapperPlugin once upstream has completed the migration.
     protected Collection<Class<? extends Plugin>> getPlugins() {
-        return Collections.singletonList(TestPlugin.class);
+        return Arrays.asList(LtrQueryParserPlugin.class, TestGeoShapeFieldMapperPlugin.class);
     }
 
     /**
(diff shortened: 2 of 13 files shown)


thePanz:
@nathancday any input or help here? :)

nomoa:
@thePanz thanks for the PR!

for createComponents hopefully the change upstream is still compatible with what we need:
```java
    @Override
    public Collection<Object> createComponents(Client client,
                                               ClusterService clusterService,
                                               ThreadPool threadPool,
                                               ResourceWatcherService resourceWatcherService,
                                               ScriptService scriptService,
                                               NamedXContentRegistry xContentRegistry,
                                               Environment environment,
                                               NodeEnvironment nodeEnvironment,
                                               NamedWriteableRegistry namedWriteableRegistry,
                                               IndexNameExpressionResolver indexNameExpressionResolver,
                                               Supplier<RepositoriesService> repositoriesServiceSupplier) {
        clusterService.addListener(event -> {
            for (Index i : event.indicesDeleted()) {
                if (IndexFeatureStore.isIndexStore(i.getName())) {
                    caches.evict(i.getName());
                }
            }
        });
        return asList(caches, parserFactory);
    }
```

For LoggingFetchSubPhaseTests I think that adding null to the metaFields argument
e.g.
```java
                    SearchHit hit = new SearchHit(doc+context.docBase, id,
                            new Text("text"), random().nextBoolean() ? new HashMap<>() : null, null);
```
should be sufficient.

Thanks!

thePanz:
@nomoa fixed the two methods, but the tests are still failing: see https://pastebin.com/5Yneh9LV

nomoa:
> @nomoa fixed the two methods, but the tests are still failing: see https://pastebin.com/5Yneh9LV

I think upstream is moving the geo_shape mapper as separate module and thus not usable in tests. Mitigation is to load a test plugin they've added.

Everytime you see a test failing try to find the `getPlugins` method of this test and change it to (e.g. for `LtrQueryBuilderTests`):
```java
    protected Collection<Class<? extends Plugin>> getPlugins() {
        return Arrays.asList(LtrQueryParserPlugin.class, TestGeoShapeFieldMapperPlugin.class);
    }
```

We'll remove this again once upstream has completed the migration.

thePanz:
Thanks for your inputs @nomoa!
Was able to get to 1 failure and 1 exception : https://pastebin.com/LqVd9evd

The exception is thrown when calling an abstract method, I guess we're missing some breaking-changes from the upstream

src/test/java/com/o19s/es/ltr/query/StoredLtrQueryBuilderTests.java
```diff
@@ -65,8 +66,9 @@
 public class StoredLtrQueryBuilderTests extends AbstractQueryTestCase<StoredLtrQueryBuilder> {
     private static final MemStore store = new MemStore();
 
+    // TODO: Remove the TestGeoShapeFieldMapperPlugin once upstream has completed the migration.
     protected Collection<Class<? extends Plugin>> getPlugins() {
-        return Collections.singletonList(TestPlugin.class);
+        return Arrays.asList(LtrQueryParserPlugin.class, TestGeoShapeFieldMapperPlugin.class);
```
nomoa, line 71: Here you need to keep the custom TestPlugin otherwize it'll fail with the message you reported about the call to the abstract method.

gradle.properties
```diff
@@ -1,2 +1,3 @@
 ltrVersion = 1.3.0
-elasticsearchVersion = 7.7.1
+elasticsearchVersion = 7.8.0
+luceneVersion = 8.5.0
```
nomoa, line 3: this is 8.5.1 (ref: https://github.com/elastic/elasticsearch/blob/v7.8.0/buildSrc/version.properties)

peterdm:
Travis build failing @nathancday.  If these first two issues are any indication, exit 56 happens when jdk.java.net is unreachable.

https://github.com/sormuras/bach/issues?q=is%3Aissue+56

Give'er another kick maybe?

```
Installing openjdk14
$ export JAVA_HOME=~/openjdk14
$ export PATH="$JAVA_HOME/bin:$PATH"
$ ~/bin/install-jdk.sh --target "/home/travis/openjdk14" --workspace "/home/travis/.cache/install-jdk" --feature "14" --license "GPL" --cacerts
Ignoring license option: GPL -- using GPLv2+CE by default
install-jdk.sh 2020-06-02
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk14" --workspace "/home/travis/.cache/install-jdk" --feature "14" --license "GPL" --cacerts" failed and exited with 56 during .
Your build has been stopped.
```

nathancday:
Zip posted to S3.
https://s3.amazonaws.com/es-learn-to-rank.labs.o19s.com/ltr-1.3.0-es7.8.0.zip
Thanks again for everything!

peterdm:
Thank you @thePanz, @nomoa and @nathancday !!!

PR #316, opened 2020-07-01
Use GitHub actions for tests and release

thePanz:
Todo
- [x] Trigger tests with actions
- [x] Trigger release with actions (when  tagging a release)
- [x] Remove usage of TravisCI

How it works:
 * two Github workflows are defined, one for running the tests, the second one to build and tag a release
 * the `tests` workflow is triggered when a PR is opened agains `master` branch, running `gradlew clean check`
 * the `release` workflow is triggered when a tag ( in the `v*.*.*` format) is pushed to the repository, it will:
   *  build the plugin (gradlew build)
   * create a release, based on the tag
   * add the ZIP file to the release (see: https://github.com/thePanz/elasticsearch-learning-to-rank/releases)


nathancday:
This looks sweet! I'll try to add a step that upload the zip to our S3  bucket too

thePanz:
@nathancday please note that the version is not containing the ES version it is build for.
How do you maintain the buils+ES compatibility?

nathancday:
Thanks for flagging that. And just to be clear this is awesome as is, thank you.

Right now tracking the ES compatibility is all manual. And we label the releases by appending the ES version to the package version in the zip file name, ltr-1.3.0-es-7.7.1.zip. Since Elastic requires all plugins to track on version numbers, it's always felt like duplication to have dueling version numbers when really the ES version *should* to capture all of the plugin progress too. 

All that to say, maybe we move to tagging based on the ES version only.

thePanz:
@nathancday thank yo for your comment!

I still like having the plugin version number tagged, to show the status of new features etc.
We could append the ES version to the asset uploaded to a version, thus marking visible the supported ES version.

The problematic bit here is: can we instruct Gradle to generate the build files, without versions (as it is now?) it would be easier to move the files and upload the m to the release :)

thePanz:
Any updates @nathancday ?

nathancday:
I haven't come back to it since the holiday, I'll schedule a review and get back next week.

nathancday:
Looks great, thanks @thePanz for all of you contributions to this project!

PR #327, opened 2020-08-31
reroute and remove demo

nathancday:


softwaredoug:
I think you meant demo in the PR title, right? That seems to be what's in the PR - and I think your intention is to redirect people to Hello LTR?

nathancday:
Just wanted to play make everyone play on "Legendary" model...Yes that's  a typo should be "demo"

PR #330, opened 2020-09-13
Add support for query time parameters for ScriptFeature term statistics

worleydl:
The term statistic injection functionality on ScriptFeatures wasn't very friendly towards query time parameters. This update will check the query parameters as a fallback if nothing is set in the script parameters.


worleydl:
I also added the sum aggregation and match/unique term counts.

PR #343, opened 2020-11-16
v7.10.0

nathancday:
Ready to review/test. There were several ES deprecations that required re-working query fetch code.


src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
```diff
@@ -32,35 +32,35 @@
...
-        // This way we reuse existing code to advance through multiple scorers/iterators
         BooleanQuery.Builder builder = new BooleanQuery.Builder();
         List<HitLogConsumer> loggers = new ArrayList<>();
         Map<String, Query> namedQueries = context.parsedQuery().namedFilters();
-        if (!(context instanceof InnerHitsContext.InnerHitSubContext)) {
+
+
+        //TODO: Need help verifying the type of the context, SearchContext was moved inside of FetchContext
+        // context instanceof InnerHitsContext.InnerHitSubContext seems to no longer be accessible
+        if (namedQueries.size() > 0) {
```
worleydl, line 63: @nomoa I can't check the instance of the SearchContext anymore as it's wrapped privately in a FetchContext.  Any ideas?  The check on namedQueries size makes the tests happy but I don't know if that catches all cases.

src/main/java/com/o19s/es/ltr/logging/LoggingFetchSubPhase.java
```diff
@@ -74,54 +74,81 @@ public void hitsExecute(SearchContext context, SearchHit[] hits) throws IOExcept
...
+            public void setNextReader(LeafReaderContext readerContext) {
+            }
+           
+            @Override
+            public void process(HitContext hitContext) throws IOException {
+                doLog(builder.build(), loggers, context.searcher(), hitContext.hit());
+            }
+        };
     }
+    // public void hitsExecute(SearchContext context, SearchHit[] hits) throws IOException {
```
nathancday, line 88: This should be ripped out before merge, but left as an artifact of the way it was before v7.10

PR #349, opened 2020-12-18
Better error handling when fetching feature/featuresets

nomoa:
NPE is currently thrown where referencing a featureset.
Throw IAE with a meaningful error message instead.


nathancday:
Thanks for adding this, I'm sure it will help folks. Also thanks for the rework of the leaf reader context for es7.10

PR #362, opened 2021-04-12
Es 7 11

worleydl:


src/main/java/com/o19s/es/ltr/action/TransportListStoresAction.java
```diff
@@ -61,20 +61,14 @@
...
     @Inject
-    public TransportListStoresAction(Settings settings, TransportService transportService,ClusterService clusterService,
-                                     ThreadPool threadPool, ActionFilters actionFilters,
-                                     IndexNameExpressionResolver indexNameExpressionResolver, Client client) {
+    public TransportListStoresAction(Settings settings, TransportService transportService,
+                                     ClusterService clusterService, ThreadPool threadPool, ActionFilters actionFilters,
+                                     IndexNameExpressionResolver indexNameExpressionResolver, String executor, Client client) {
         super(ListStoresAction.NAME, transportService, clusterService, threadPool,
-            actionFilters, ListStoresActionRequest::new, indexNameExpressionResolver);
+            actionFilters, ListStoresActionRequest::new, indexNameExpressionResolver, ListStoresActionResponse::new, ThreadPool.Names.SAME);
```
worleydl, line 68: We may want to use the new `executor` param.

nathancday: Whoops didn't mean to resolve this, I has the swapped in for ThreadPool at some point, but I will put `executer` back in there tonight

worleydl: I tried passing it thru and it caused tests to fail out so I'm sticking with the SAME for now.

worleydl:
I think this is ready for prime time.  We may want to do some cleaning of the commented out lines in build.gradle but otherwise 👍

nathancday (approved):
Damn that green check feels good!

PR #372, opened 2021-06-16
Add support for 7.13.2

JorisAndrade:
After https://github.com/o19s/elasticsearch-learning-to-rank/pull/371 I guess this is all what is needed for a 7.13.2 build.

Let me know if this PR is missing something.

PR #374, opened 2021-06-16
Joris andrade joris/es 7 13 2

nathancday:
replacing #372 with this just b/c when I pushed a small change it made a new branch instead of adding to Joris's original....


JorisAndrade:
> replacing #372 with this just b/c when I pushed a small change it made a new branch instead of adding to Joris's original....

No worries thanks!

Issue #375, opened 2021-06-20
match_explorer query: some problems with IndexReader.numDocs() while calculating idf


ndkmath1:
Now the plugin is using this way to calculate idf: https://github.com/o19s/elasticsearch-learning-to-rank/blob/main/src/main/java/com/o19s/es/explore/ExplorerQuery.java#L111
`idf_stats.add(sim.idf(tStats.docFreq(), searcher.getIndexReader().numDocs()));`

There're three problems

1. TermStatistics.docFreq() counts deleted document but IndexReader.numDocs() doesn't  (link: https://lucene.apache.org/core/8_9_0/core/org/apache/lucene/search/TermStatistics.html.  "Values may include statistics on deleted documents that have not yet been merged away.")
In Lucene, to calculate idf, "...CollectionStatistics.docCount() is used instead of IndexReader#numDocs()" (link: https://lucene.apache.org/core/8_9_0/core/org/apache/lucene/search/similarities/TFIDFSimilarity.html#idfExplain-org.apache.lucene.search.CollectionStatistics-org.apache.lucene.search.TermStatistics-)

2. numDocs() doesn't depend on search_type (query_then_fetch and dfs_query_then_fetch), it always counts on shard level for both search_type but docFreq() depends on search_type
- if search_type is dfs_query_then_fetch, docFreq will count on index level
- if search_type is query_then_fetch, docFreq will count on shard level

Therefore, number of documents should also depend on search_type.

3. if mappings has nested field, numDocs() will count hidden documents. Imagine that my shard has 1 document and this document has a nested field for arrays of 100 objects. Then numDocs() is equal 101. (link: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html)

These three problems will be solved if we do as follows (I can help with a pull request if it makes sense)
`idf_stats.add(sim.idf(tStats.docFreq(), searcher.collectionStatistics(term.field()).docCount()));`

Do the problems and the way to solve it make sense? Let me know if I need to clarify anything.

KhanhNguyen6:
> 2\. docFreq() depends on search_type
> 
> * if search_type is dfs_query_then_fetch, docFreq will count on shard level
> * if search_type is query_then_fetch, docFreq will count on index level

I had a small typo in (2). It should be
- if search_type is dfs_query_then_fetch, docFreq will count on index level
- if search_type is query_then_fetch, docFreq will count on shard level
I can provide a small example about this problem if needed.

[...]

PR #378, opened 2021-06-22
workaround for issue #375

ndkmath1:


nathancday:
Thanks this helps me see it in action.

worleydl:
Thanks! I applied the same fix to the TermStatQuery and added a test.  If no one has complaints we'll get this merged up and cut into a new release.

ndkmath1:
Thank you!

Issue #356, opened 2021-01-28
`min_raw_df` returns `Float.MAX_VALUE` for non-vocabulary words


wxd:
ElasticSearch version: `7.8.0` or `7.9.0` (in Docker)
Plugin version: `ltr-1.3.0-es7.8.0` or `ltr-plugin-v1.5.0-es7.9.0`

Are there any solutions/workarounds?

#### `min_raw_df`
```json
{"query": {"match_explorer": {"type": "min_raw_df", "query": {"match": {"text": "xxx"}}}}}
```

```json
{"hits": {"max_score": 3.4028235E38}}
```

#### `max_raw_df`
```json
{"query": {"match_explorer": {"type": "max_raw_df", "query": {"match": {"text": "xxx"}}}}}
```

```json
{"hits": {"max_score": 0.0}}
```

worleydl:
Hello,

This looks like a bug.  I will test this week and let you know.

worleydl:
Sorry for the wait, this will be fixed with PR #378

worleydl:
Well it's partially fixed, I think collection stats are okay but document specific still needs some work.  `min_raw_tp` for example is still surfacing the behavior.

worleydl:
After testing some more the doc specific stats have the following behaviors:
- If the query has no vocab words a value of 0 will be returned
- If the query has a mix of vocab and non, the non-vocabulary words will be excluded from the statistics.  This means a min(vocabWord, nonVocabWord) will always return the vocabWord and never 0.

This is due to the internal design of match explorer query.  From testing the term stat query things behave a little more predictably at the cost of some performance.  For now I think this particular issue is solved.

PR #380, opened 2021-06-24
Doc freq issue

ndkmath1:


src/main/java/com/o19s/es/explore/ExplorerQuery.java
```diff
@@ -104,11 +105,11 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
 
             for (Term term : terms) {
                 TermStates ctx = TermStates.build(searcher.getTopReaderContext(), term, scoreMode.needsScores());
-
-                if(ctx != null){
-                    df_stats.add(ctx.docFreq());
-                    idf_stats.add(sim.idf(ctx.docFreq(), searcher.collectionStatistics(term.field()).docCount()));
-                    ttf_stats.add(ctx.totalTermFreq());
+                if(ctx != null && ctx.docFreq() > 0){
+                    TermStatistics tStats = searcher.termStatistics(term, ctx.docFreq(), ctx.totalTermFreq());
```
worleydl, line 109: I dug into TermStatistics yesterday and it appears to just be a wrapper around the values you pass in for docFreq and ttf.  Can you provide an example where we would get different values from using `TermStatistics::docFreq` over using `TermStates::docFreq`?

worleydl:
Apologies if I broke your last PR trying to get a fix in for #356 at the same time.  Currently there's a problem around the StatisticsHelper if you never put any data into them.  This can be fixed by altering the helper to handle cases where no data entered, or we could add an else to line 108 and set all the values as 0

src/main/java/com/o19s/es/explore/ExplorerQuery.java
ndkmath1: I guess the problem is` TermStatistics.termStatistics` can be [overridden](https://lucene.apache.org/core/8_9_0/core/org/apache/lucene/search/IndexSearcher.html#termStatistics-org.apache.lucene.index.Term-int-long-). Lucene says: "This can be overridden for example, to return a term's statistics across a distributed collection." 
and that behavior doesn't happen to `TermStates`. I will provide an example in a bit.

worleydl: Okay I see one example at https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/search/dfs/DfsPhase.java#L37 which makes sense.  

If it's not a pain it would be interesting to get tests built around these cases and to possibly support both shard/index level stats via a param.

ndkmath1: Hey Dan,
I provided an reproducible example [here](https://github.com/o19s/elasticsearch-learning-to-rank/issues/381).

ndkmath1: I also see an "overridden" example here: https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/search/internal/ContextIndexSearcher.java#L279

ndkmath1:
I see TermStatQuery also use TermStates: https://github.com/o19s/elasticsearch-learning-to-rank/blob/main/src/main/java/com/o19s/es/termstat/TermStatSupplier.java#L77
I think we should apply the same change as the PR for TermStatQuery.

src/main/java/com/o19s/es/explore/ExplorerQuery.java
worleydl: Makes sense, I've updated the TermStatSupplier as well.

src/main/java/com/o19s/es/explore/ExplorerQuery.java
```diff
@@ -106,10 +112,16 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
...
+                    if (shard) {
+                        df_stats.add(ctx.docFreq());
+                        idf_stats.add(sim.idf(ctx.docFreq(), searcher.getIndexReader().getDocCount(term.field())));
+                        ttf_stats.add(ctx.totalTermFreq());
+                    } else {
+                        TermStatistics tStats = searcher.termStatistics(term, ctx.docFreq(), ctx.totalTermFreq());
+                        df_stats.add(tStats.docFreq());
+                        idf_stats.add(sim.idf(tStats.docFreq(), searcher.collectionStatistics(term.field()).docCount()));
+                        ttf_stats.add(tStats.totalTermFreq());
+                    }
```
ndkmath1, line 124: Hey Dan,
I have a question. Why do you need if/else here? I think we just need to use TermStatistics and everything will work correctly afterwards.
I mean if we just use TermStatistics (no need if/else) and do a request like below
```
POST my_index/_search?search_type=dfs_query_then_fetch
{
   "query": {
       "match_explorer": {
           "type": "sum_classic_idf",
           "query": {
               "match": {
                   "title": "when breath becomes air"
               }
           }
       }
   }
}
```
Both `docCount` and `docFreq` are over index.

If do a request like below (default if no `search_type`, then `search_type=query_then_fetch`)
```
POST my_index/_search
{
   "query": {
       "match_explorer": {
           "type": "sum_classic_idf",
           "query": {
               "match": {
                   "title": "when breath becomes air"
               }
           }
       }
   }
}
```
Both `docCount` and `docFreq` are over shard

worleydl: You're right. Cleanup inbound.

I'm having trouble getting the behavior to work correctly in the term stat query, I'm slightly unfamiliar with the `dfs_query_then_fetch` lifecycle so maybe something is incompatible.

ndkmath1: I don't know much about the code of term stat query. Hopefully, you will figure out something  with `dfs_query_then_fetch` for term stat query soon.

PR #394, opened 2021-10-05
Fix exception in LoggingSearchExtBuilder JSON

thePanz:
this PR solves the following exception, thrown when trying to build the Ext property for ES

```
org.elasticsearch.ElasticsearchException: com.fasterxml.jackson.core.JsonGenerationException: Can not start an object, expecting field name (context: Object)

Caused by: com.fasterxml.jackson.core.JsonGenerationException: Can not start an object, expecting field name (context: Object)
[..]
at com.o19s.es.ltr.logging.LoggingSearchExtBuilder.toXContent(LoggingSearchExtBuilder.java:80) ~[elasticsearch-learning-to-rank-1.5.7-es7.13.4.jar:1.5.7-es7.13.4]
[..]
```

The current fix I used in one of my projects is to create a new class:
```java

import java.io.IOException;
import org.elasticsearch.common.xcontent.XContentBuilder;

/**
 * The original implementation is affected by a bug, and it is building an invalid JSON.
 */
public final class LoggingSearchExtBuilderFix extends com.o19s.es.ltr.logging.LoggingSearchExtBuilder {
	@Override
	public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
		builder.field(NAME);
		return super.toXContent(builder, params);
	}
}
```

But I would prefer a fix in the original library :)

### DIFF
--- a/src/main/java/com/o19s/es/ltr/logging/LoggingSearchExtBuilder.java
+++ b/src/main/java/com/o19s/es/ltr/logging/LoggingSearchExtBuilder.java
@@ -77,7 +77,7 @@ public class LoggingSearchExtBuilder extends SearchExtBuilder {
 
     @Override
     public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
-        builder.startObject();
+        builder.startObject(NAME);
         builder.field(LOG_SPECS.getPreferredName(), logSpecs);
         return builder.endObject();
     }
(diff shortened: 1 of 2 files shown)


thePanz:
Well, the test failure is weird ... as I am currently using the proposed fix on live project! :confused:

nathancday:
Agreed, the build fail doesn't make any sense. Might be something with the versions. Thanks for sharing this!

nomoa:
@thePanz could you provide reproducible steps?

I'm not sure the search_ext name has to be provided here since it's being given when declaring the extension itself: https://github.com/o19s/elasticsearch-learning-to-rank/blob/main/src/main/java/com/o19s/es/ltr/LtrQueryParserPlugin.java#L161.

The test is failing because you changed the way the logging extension is being serialized as json without changing how it parsed and this was captured by the test.

thePanz:
ping @hhassan1

worleydl:
Thanks for the update and test fix! Merging and will be part of the 1.5.8 release.

PR #412, opened 2022-01-19
Upgrade to elastic 6.8.23

iemarjay:


nathancday:
Hey Joseph. Is this a release that you need? I'd like to know you
motivation, because if so we could figure out another way to get GH actions
to build it for you. I really like your idea of supporting the LTS releases
from version 6.

PR #437, opened 2022-09-13
Branch es 8.5 wrapped

grcevski:
Example conversion to using the new lang-expression facility in ES 8.5.0.


worleydl:
This should be good to go.  We will need to update `build.gradle` and `gradle.properties` with the final builds that include the updates from this PR: https://github.com/elastic/elasticsearch/pull/89895

grcevski:
Hi @worleydl, we've merged https://github.com/elastic/elasticsearch/pull/89895, so it will make it in 8.5.0 which should go into feature freeze soon. I'll let you decide if you want to merge this into your branch or pick and choose parts of it.

PR #452, opened 2023-01-03
xgboost to handle missing values

lechipatrick:
Currently, the `NaiveAdditiveDecisionTree` class assumes that all feature values are not missing. In certain applications, missing feature values are not easily addressed (it might be hard to decide what appropriate value to fill in for missing). 

This PR enables `NaiveAdditiveDecisionTree` to correctly score features even when some features have a missing value.

Related issue: https://github.com/o19s/elasticsearch-learning-to-rank/issues/135

### DIFF
--- a/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
+++ b/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java
@@ -94,12 +94,18 @@ public class NaiveAdditiveDecisionTree extends DenseLtrRanker implements Account
         private final Node right;
         private final int feature;
         private final float threshold;
+        private final int leftNodeId;
+        private final int rightNodeId;
+        private final int missingNodeId;
 
-        public Split(Node left, Node right, int feature, float threshold) {
+        public Split(Node left, Node right, int feature, float threshold, int leftNodeId, int rightNodeId, int missingNodeId) {
             this.left = Objects.requireNonNull(left);
             this.right = Objects.requireNonNull(right);
             this.feature = feature;
             this.threshold = threshold;
+            this.leftNodeId = leftNodeId;
+            this.rightNodeId = rightNodeId;
+            this.missingNodeId = missingNodeId;
         }
 
         @Override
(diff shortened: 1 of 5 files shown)


nathancday:
I think this looks great, thanks @lechipatrick and sorry for leaving this since January.

My only little nit: I don't think we need `rightNodeId` as its never checked in code, so we could save that byte.

@worleydl can you tell me if I'm wrong?

patrick-le-shipt:
> I think this looks great, thanks @lechipatrick and sorry for leaving this since January.
> 
> My only little nit: I don't think we need `rightNodeId` as its never checked in code, so we could save that byte.
> 
> @worleydl can you tell me if I'm wrong?

Thanks, @nathancday and good catch! I'll let @worleydl chime in first before addressing.

styrmis:
In the models we are training we see that the missing node specified in the model is sometimes the left, and other times the right. With the current implementation we are therefore only following the correct path coincidentally in some cases, but not others, which causes our model to behave differently in production than in the training process.

It would be great to see this merged—is there anything that can be done here to help, e.g. testing? Happy to contribute where possible.

sstults:
@patrick-le-shipt if you agree with @nathancday then go ahead and make that change.

wrigleyDan:
@patrick-le-shipt @lechipatrick @styrmis 
I removed `rightNodeId`, local tests run successful. Would any of you care to take a look to make sure I didn't miss anything?

styrmis:
@wrigleyDan I've taken a look and haven't found any issues, though I'm finding [dectree/simple_tree.txt](https://github.com/o19s/elasticsearch-learning-to-rank/compare/main...xgboost_missing_values#diff-97e55673ffc610350e83955d39900714c9b1abc3f9d456bc0f453a042c386f2a) quite hard to follow as it references node ids, but the file itself doesn't contain ids for the nodes.

Still though, haven't found any actual issues, looking forward to trying this out 👍

wrigleyDan:
Thanks for review @styrmis.
Created a separate PR (https://github.com/o19s/elasticsearch-learning-to-rank/pull/480) for this and merged into main, so I'm closing this one.
Thanks @lechipatrick for this contribution.
I will cut a fresh release soon with this new feature improvement.

styrmis:
To add some further context, I profiled one of our models and in terms of cases where the missing branch matches either the yes or no node, got:

Yes count: 33283, No count: 27104

My understanding is that the [left node is the yes node](https://github.com/o19s/elasticsearch-learning-to-rank/blob/05ad18c44af7a3dcb200ded8a6584ecbe1367bda/src/main/java/com/o19s/es/ltr/ranker/parser/XGBoostJsonParser.java#L158), and that the yes/true case is when [the score is less than the threshold](https://github.com/o19s/elasticsearch-learning-to-rank/blob/b9aecbebabc4199cfc2eebcdbba43518e99d9e16/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java#L116). If comparing to `NaN` then we will always choose the no/right branch, which accounts for fewer than half of the missing node assignments in our model.

PR #471, opened 2023-08-08
Update es7_17_X to Elasticsearch 7.17.11

hexa2k9:
We're looking into usage of this plugin but we'd like to install a matching Version on our Clusters.


hexa2k9:
Hey @nathancday,

any chance you can look into this given you merged #466 as well?

Thanks in advance.

PR #479, opened 2023-12-08
Add rank:ndcg and rank:map as valid objectives

styrmis:
When attempting to use `rank:ndcg` as an objective model deployment failed unexpectedly, stating that the objective was invalid.

As per the [XGBoost docs](https://github.com/dmlc/xgboost/blob/1094d6015df7be6f5829e53a4c257f3481b43ee5/doc/parameter.rst#parameters-for-learning-to-rank-rank-ndcg-rank-map-rank-pairwise) `rank:ndcg` and `rank:map` are valid objectives, but are currently rejected by the parser.

This change resolves this by extending the list of objectives which map to the no-op normalizer, allowing the models to be parsed and used by the plugin.


wrigleyDan:
LGTM, merging with main.

PR #480, opened 2023-12-14
Update - xgboost to handle missing values 

wrigleyDan:
This PR is an updated version of https://github.com/o19s/elasticsearch-learning-to-rank/pull/452

Work was done by @lechipatrick, improvement spotted by @nathancday, improvement reviewed by @styrmis

With this change XGBoost can handle missing feature values in `NaiveAdditiveDecisionTree`.


styrmis:
@wrigleyDan Thanks again for getting this merged—do we know when it might appear in a release? It looks like the last release was a few days before this was merged.

wrigleyDan:
@styrmis I wanted to release this together with upgrading the plugin to ES 8.11.x.
The latter proves to be less straight forward than other upgrades but we'll hopefully resolve that soon. If not we can cut a release for the latest improvements to not make that too dependent on the upgrade.

styrmis:
Thanks @wrigleyDan. We have been looking again at this issue of inference parity between the plugin and XGBoost and have discovered that as the [`NaiveAdditiveDecisionTree` is a `DenseLtrRanker`](https://github.com/o19s/elasticsearch-learning-to-rank/blob/05ad18c44af7a3dcb200ded8a6584ecbe1367bda/src/main/java/com/o19s/es/ltr/ranker/dectree/NaiveAdditiveDecisionTree.java#L31), it fills in missing values with 0. As such the only way currently that we can achieve parity in terms of inference between training and Elasticsearch/production is to fill in missing values in the training data with 0 also.

I've [opened an issue](https://github.com/o19s/elasticsearch-learning-to-rank/issues/481) for this.

PR #503, opened 2024-11-27
chore: bumps to elastic 8.15.5

cdmbr:


cdmbr:
ping @Philippus @wrigleyDan

Philippus:
Not sure about ignoring a test, but other than that lgtm. I don't have any maintenance rights in this repo though.

wrigleyDan (approved):
Sorry for leaving this unattended for so long, it completely slipped from my list.

I think disabling the test is alright for now. I'd rather have a fix for it but I don't see that coming any time soon.

Going ahead and merging this.

PR #524, opened 2026-01-28
Update to Elasticsearch 9.2.4 / Lucene 10.3.2

CalderLund:
This change bumps the plugin’s build target from Elasticsearch 9.0.8 to 9.2.4 (and Lucene 10.1.0 -> 10.3.2) so the plugin can be installed on and run against ES 9.2.4 (plugins must match ES major.minor.patch).

- Elasticsearch plugins are binary/version coupled; building against 9.0.8 will not load on 9.2.4.
- ES 9.2.x + Lucene 10.3.x introduced API and serialization changes that required code updates.

- Version bump:
  - `gradle.properties`: `elasticsearchVersion=9.2.4`, `luceneVersion=10.3.2`
- ES API compatibility updates:
  - `ActionPlugin.ActionHandler` is non-generic in ES 9.2; update `LtrQueryParserPlugin#getActions()` to use the new signature/type.
  - Transport request type changes: node-level requests now extend `AbstractTransportRequest` where ES 9.2 no longer uses `TransportRequest` as a base class for these usages.
  - Response stream constructors: remove `super(StreamInput)` calls from direct `ActionResponse` subcsses because ES 9.2’s `ActionResponse` no longer has a `StreamInput` constructor (node/node(s) responses continue to call `super(in)` via `BaseNodeResponse`/`BaseNodesResponse`).
  - Metadata API: adapt index lookups to project-scoped metadata (`metadata().getProject().index(...)`) to match ES 9.x model changes.
  - SearchHit API: update document field logging to the new `SearchHit#setDocumentField(DocumentField)` signature.
  - Lucene API: replace `new DisiPriorityQueue(size)` with `DisiPriorityQueue.ofMaxSize(size)` for Lucene 10.3.
  - Tests: update removed constants/constructors (e.g., TransportVersions constant removal; IndexNameExpressionResolver constructor parameter changes).
- Fix a latent deserialization bug exposed by the upgrade:
  - `TransportClearCachesAction#newNodeResponse(...)` now returns `new ClearCachesNodeResponse(in)` instead of `null`, which is required for proper node response deserialization under ES 9.2.

- ES 9.x test framework can optionly enable entitlement instrumentation during `ESTestCase` bootstrap via system properties.
- In this plugin repo’s Gradle unit-test JVM (`./gradlew test`), enabling entitlements without a full ES distro-style bootstrap can fail before tests run.
- We therefore default `es.entitlement.enableForTests=false` for Gradle `Test` tasks *only when the property is not already set*, preserving an opt-in path (`-Des.entitlement.enableForTests=true`) for environments that can correctly provide the required entitlement bootstrap wiring.
- This does not affect plugin runtime behavior in a real Elasticsearch distribution; it only affects the unit-test JVM bootstrap path.

- Ran with JDK 21:
  - `./gradlew test`
  - `./gradlew clean test`
  - `./gradlew clean check` (includes unit tests, `javaRestTest`, `yamlRestTest`, checkstyle, and packaging checks)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,3 +87,14 @@ java {
 // Set to false to not use elasticsearch checkstyle rules
 checkstyleMain.enabled = true
 checkstyleTest.enabled = true
+
+// Elasticsearch test-framework (9.x) can enable the entitlement agent via system properties.
+// When enabled, it needs either `es.entitlement.agentJar` or an ES-style `es.path.home` that
+// contains `lib/entitlement-agent/*.jar`. In this plugin repo, tests run without a distro
+// layout. If entitlements are enabled without the required wiring, ES test framework
+// bootstrap fails before any tests run. Default to disabling unless explicitly enabled.
+tasks.withType(Test).configureEach {
+  if (System.getProperty("es.entitlement.enableForTests") == null) {
+    systemProperty "es.entitlement.enableForTests", "false"
+  }
+}
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
-ltrVersion = 1.5.12
-elasticsearchVersion = 9.0.8
-luceneVersion = 10.1.0
+ltrVersion = 1.5.13
+elasticsearchVersion = 9.2.4
+luceneVersion = 10.3.2
 ow2Version = 8.0.1
 antlrVersion = 4.5.1-1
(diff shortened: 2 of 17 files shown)


wrigleyDan:
Thanks for this PR and sorry for not responding earlier, that slipped through. I'll have a look later this week to move this forward 👀

wrigleyDan (approved):
LGTM, thanks!

